PR #15, opened 2021-11-25
Support darwin-amd64

dlvhdr:
There was a bug in the release script that didn't publish the darwin-amd64 exe


dlvhdr:
@abdfnx I fixed it, but can you make sure the other exes are configured ok?
The amd64 one was easy for me to test since that's what I'm using.

abdfnx:
ok I will test it all

PR #17, opened 2021-11-25
rename darwin 64 base name to `darwin-amd64` and windows

abdfnx:


abdfnx:
can you now release a new one ?

Issue #22, opened 2021-12-06
Can the list of repos be optional?


KyleKing:
It looks like specifying the repos is required from my local testing. i wanted to match the default Github PRs tab with the below snippet, but no PRs were matched

```yaml
- title: Created
  filters: author:@me
- title: Assigned
  filters: assignee:@me
- title: Mentioned
  filters: mentions:@me
- title: Reviewed Requested
  filters: review-requested:@me
```

I confirmed that specifying one or more repos does work though:

```yaml
- title: Created
  repos:
    - KyleKing/commitizen
    - timothycrosley/pdocs
  filters: author:@me
```

I assume this would require pulling a list of repos if the repos list is empty here: https://github.com/dlvhdr/gh-prs/blob/37e60a8409b2ba5c46d52e9ea00d5a94cbaa2d73/ui/section.go#L43

tkinz27:
I would love the equivalent of 

```
https://github.com/pulls?q=user%3ASomeOrgName
```

dlvhdr:
Hey @KyleKing  @tkinz27, I would love for you to join the discussion in #27.
It's relevant to this issue and would love some more input.

KyleKing:
I was playing around with the [Github Explorer](https://docs.github.com/en/graphql/overview/explorer) based on [your example](https://github.com/dlvhdr/gh-prs/pull/27#issuecomment-991019903) and generated a unique list of all my repositories that have open PRs which I added to my configuration manually!

```sh
RES=$(gh api graphql --paginate -f query='query ($endCursor: String) {
  viewer {
    login
    pullRequests(first: 100, after: $endCursor, states: OPEN) {
      nodes {
        title
        url
        baseRepository {
          name
          nameWithOwner
          owner {
            login
          }
        }
        headRepository {
          name
          nameWithOwner
          owner {
            login
          }
        }
        number
      }
    }
  }
}')
echo $RES | jq '.data.viewer.pullRequests.nodes[].baseRepository.nameWithOwner' | uniq
```

dlvhdr:
yeah that's pretty much the query we'll need to add :)

dlvhdr:
@KyleKing  @tkinz27  See https://github.com/dlvhdr/gh-prs/pull/40
I'm moving to only use GitHub filters so from the next release you can omit the Repos key and specify filters which allow everything [github.com/pulls]() allows

PR #27, opened 2021-12-09
Support checking out PRs

barakwei:
This PR adds support for checking out the selected PRs using gh.

There's room for improvements:
* Better handling when the current repo is unrelated to the selected PR. Right now there's no handling.
* Force checkout support.﻿

### DIFF
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -77,7 +77,6 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		cmd  tea.Cmd
 		cmds []tea.Cmd
 	)
-
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
 		switch {
@@ -119,6 +118,15 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			currPR := m.getCurrSection().Prs[m.cursor.currPrId]
 			utils.OpenBrowser(currPR.Url)
 			return m, nil
+
+		case key.Matches(msg, m.keys.Checkout):
+			currSection := m.getCurrSection()
+			if currSection.numPrs() == 0 {
+				return m, nil
+			}
+			currPR := m.getCurrSection().Prs[m.cursor.currPrId]
+			utils.CheckoutPR(currPR.Number)
+			return m, tea.Quit
 
 		case key.Matches(msg, m.keys.Help):
 			m.help.ShowAll = !m.help.ShowAll
--- a/utils/keys.go
+++ b/utils/keys.go
@@ -6,6 +6,7 @@ type KeyMap struct {
 	Up          key.Binding
 	Down        key.Binding
 	Open        key.Binding
+	Checkout    key.Binding
 	PageDown    key.Binding
 	PageUp      key.Binding
 	NextSection key.Binding
@@ -23,7 +24,7 @@ func (k KeyMap) FullHelp() [][]key.Binding {
 		{k.Up, k.Down},
 		{k.PrevSection, k.NextSection},
 		{k.PageDown, k.PageUp},
-		{k.Open},
+		{k.Open, k.Checkout},
 		{k.Help, k.Quit},
 	}
 }
@@ -56,6 +57,10 @@ var Keys = KeyMap{
 	Open: key.NewBinding(
 		key.WithKeys("o"),
 		key.WithHelp("o", "open in GitHub"),
+	),
+	Checkout: key.NewBinding(
+		key.WithKeys("c"),
+		key.WithHelp("c", "checkout"),
 	),
 	Help: key.NewBinding(
 		key.WithKeys("?"),
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"math"
+	"os"
 	"os/exec"
 	"runtime"
 	"strconv"
@@ -37,6 +38,23 @@ func OpenBrowser(url string) {
 	default:
 		err = fmt.Errorf("unsupported platform")
 	}
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func CheckoutPR(id int) {
+	checkoutCmd := exec.Command("gh", "pr", "checkout", strconv.Itoa(id))
+	checkoutCmd.Stdin = os.Stdin
+	checkoutCmd.Stdout = os.Stdout
+	checkoutCmd.Stderr = os.Stderr
+
+	err := checkoutCmd.Start()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = checkoutCmd.Wait()
 	if err != nil {
 		log.Fatal(err)
 	}


dlvhdr:
This is a really nice idea @barakwei! Thanks for writing this.
I think we should definitely think about the points you made before we merge.

What's your use case? Code reviews?

barakwei:
Exactly, I may have 5-6 open PRs of my own + some to review. This really makes it easy to move between them.
For my branches, I often don't remember the branch name for each PR.
For my assigned/review-requested PR, sometimes I want to checkout the PR. Our developers often push-force to the PR, so the force checkout is important.

In general this PR doesn't match this repo, I find the config file approach as a complication, so it'll be interesting to integrate it.
Maybe when I'm in a git repo show all the PRs related to that repo by default?
Maybe we can flags that can help us?

Unrelated question - I want the tool to show all the PRs from all repos (similar to https://github.com/pulls) and not to filter by repo. Is that possible?

dlvhdr:
Yeah the use case you described is something cool we could support.
I wrote this tool because my team works with a lot of different repos all under the same organization.
Often I have PRs in multiple repos.
Also, I want to know which PRs my team members (or others across the company) write so I have more context.
So I wrote the tool get an overview of these things.

It wasn't really my intention to support a workflow where you're focusing on a certain repo.
I do see it as something this can branch out to.
I need to think what's the best way to add this.
I can imagine a default section added for when you're in a git repo and run the tool.
It will then allow checking out PRs which will require them to be pulled.
For that I think we need to add some way of displaying errors.
That's just an idea, there could be other better ways.

Regarding the second question i think this can be done.
` gh repo list` will list all repos owned by you.
`gh repo list cli` will list all repos owned by the "cli" org.
So we need to add a way to add that into the configuration.

barakwei:
OK I think I now understand the issue. I'm sure there's a technical way to get all the PRs a user created (without specifying a user), maximum we'll use graphql (using `gh api`).

What I propose is this:
1. Drop the configuration. In favor or hard-coded tabs like in https://github.com/pulls. Maybe start with PRs the user created, assigned and review required.
2. If you're in a repo, it'll show the relevant things from the repo. If you're not in a repo, it'll error (like similar `gh` commands)
3. If you provide the `--all` flag, and it'll show data from all repos.
4. If `--all` is used, you can also add `--org` and specify the org.
5. Maybe if you're in a repo, and opened the UI, there'll a be keys to change to all repos.

Will this cover your usage?

dlvhdr:
What you're suggesting will take the project to a very different direction...
My org has about 3000 repos, and I only care about ~6 of them.
I also like to follow some repos I'm interested in by the OSS community.

Now, I don't actually like [github.com/pulls](github.com/pulls). The tabs there are mostly irrelevant to me.
I never got assigned a PR (not sure what that even means).
The most relevant sections I would take from there are the "My pull requests" and "Review required" ones.

The config file allows me to arbitrarily select the repos I'm interested in.
Orgs can have lots of repos that are not relevant to me and the repos I'm interested in change over time.
So I think the configuration is a strong concept.

Let me know what you think.

So to summarize:
1. We  can add another field to the config that says where each repo is located locally then we could support checking out branches.
2. Regarding showing the branch name - this can be easily added to the table
3. We can add a "org" field to the config file that will fetch all repos in that org.
4. I need to check out the graphql api to see how this would allow not specifying any repo/org.

dlvhdr (changes requested):
Requesting changes for now, just to track this.

KyleKing:
FWIW, here are a few of my comments!

## General

I think there should be a default configuration file so that new users can get started quickly (I added a few notes in https://github.com/dlvhdr/gh-prs/issues/5#issuecomment-991393190)

## CLI ~ Configuration

I think @barakwei's ideas in https://github.com/dlvhdr/gh-prs/pull/27#issuecomment-990751703 are not necessarily mutually exclusive with the configuration-file approach and actually could be a really useful addition:

- Maybe the first few tabs are generated based on the CLI flags and/or the current directory:
    - When launched from a git directory, the open PRs for the repository are shown in the first tab
    - Any CLI flag will add a tab in the order they are passed:
        - `--org-id=SomeOrg` will add a "SomeOrg" tab
        - `--all` will add an "All" tab
        - etc.
    - Then all tabs configured in the configuration file will be shown

I think this would be a best of both approach, but I have no idea how hard that would be to implement

### Example

For an example configuration file (assuming that #22 is implemented to make repos filtering optional):

```yml
- title: Created
  filters: author:@me
- title: Mentioned
  filters: mentions:@me
- title: Reviewed Requested
  filters: review-requested:@me
- title: Subscribed
  repos:
    - cli/cli
    - charmbracelet/glamour
    - charmbracelet/lipgloss
  filters: -author:@me
```

And command:

```sh
~/some/git/dir/project_a $ gh prs --all --org-id=SomeOrg
```

The tabs would render in this order:

| project_a    | All       | SomeOrg    | Created   | Mentioned  | Reviewed Requested  | Subscribed |
|--------------|-----------|------------|-----------|------------|---------------------|------------|

## PR Checkout

For checking out PRs, I think that would be a really common workflow (I would use it!) and would be a good addition

You motivated me to go back to using single directories per PR and I put together an alias for my opionated approach (`~/developer/checkouts/<org>/<repo>/pr<number>`):

```sh
export PR_CHECKOUT_DIR=~/developer/checkouts
alias clone-pr() {
    cd $PR_CHECKOUT_DIR && gh repo clone $1 $1/$2 && cd $1/pr$2 && gh pr checkout $2 --force
}

# Use with the org/repo and PR Number
clone-pr timothycrosley/pdocs 25
```

dlvhdr:
Thanks @KyleKing!
I think what you're suggesting is the way to go as well.

dlvhdr:
So to clarity @KyleKing, the `--all` flag will show a section that shows which PRs exactly?
PRs created by you from across all repos?
If so, what's the difference between this and:
```
- title: Created
  filters: author:@me
```

KyleKing:
I was thinking about showing PRs that I've "watched." This would be the best approximation of `--all`, because it includes created, mentioned, assigned, or manually subscribed. I'm not sure how it would be implemented, but looks like it should be possible because you can see all subscriptions on GitHub (which could be filtered for PRs): https://github.com/notifications/subscriptions. I don't think the GitHub query syntax supports this right now: https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests

Your comment is a good point. Adding CLI flags won't provide any benefit other than ordering the tabs on the fly, so if anything, maybe a CLI flag to show a single tab might be a good future feature? Otherwise, the configuration file with optional repositories is the better approach

Pakon2543 (approved):

ndom91:
Why did this never get merged if I may ask?

dlvhdr:
It needed more exploration/thought. Read the discussion

dlvhdr:
I'd like to propose an "API" to checkout the current PR.
It will use [Go templates](https://pkg.go.dev/text/template).

In the `config.yml` file, you will include this:
```yaml
defaults:
  prCustomCommand: <some bash command>
```

When you press <kbd>c</kbd> I will run this command on the currently selected PR.
The command will run in a new bash process.

Using Go templates, I can provide the user with various params he can inject together with logic.

## Example with Neovim

```yaml
defaults:
  prCustomCommand: >
    tmux new-window -c
    {{if eq .RepoName "dlvhdr/gh-dash"}}~/code/personal/gh-dash{{end}}
    'gh pr checkout {{.PrNumber}} &&
    nvim -c ":DiffviewOpen master...types-service-cr"'
```

Here I'm using this to:
1. Open a new tmux window
1. Use the `gh` CLI to checkout the PR
1.  launch Neovim with the [diffview.nvim](https://github.com/sindrets/diffview.nvim/) extension to do a code review.

## Example with VSCode

You can also do the same with VSCode with:
```yaml
defaults:
  prCustomCommand: 
    code {{if eq .RepoName "dlvhdr/gh-dash"}}-n ~/code/personal/gh-dash &&
    cd ~/code/personal/gh-dash{{end}} &&
    gh pr checkout {{.PrNumber}}
```
You can then run the "GitHub Pull Request: Focus on Active Pull Request View" to do a review.

## What else?

Additionally to having the ability to checkout a PR, you can use this command to do whatever you want like:
* Copy the PR branch name
* Run a cURL command against the GitHub API
* Or whatever

## Testing
I've tested both examples and it works on Neovim and VSCode.

WDYT?

PR #39, opened 2021-12-22
Add support to dynamically fetch PR if config file not present

arush-sal:
Signed-off-by: Arush Salil <me@aru.sh>

### DIFF
--- a/config/parser.go
+++ b/config/parser.go
@@ -6,7 +6,12 @@ import (
 	"path"
 	"path/filepath"
 
+	"strings"
+
 	"gopkg.in/yaml.v2"
+
+	gh "github.com/cli/go-gh"
+	"github.com/cli/go-gh/pkg/api"
 )
 
 type SectionConfig struct {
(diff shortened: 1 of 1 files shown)


dlvhdr:
Hi @arush-sal :)
Can you please provide more information on what this pull request does?

* Please provide a screenshot or even better a video of how the new flow looks/works.
* What's the use case for this?

arush-sal:
@dlvhdr 
*What this pull request does:*
* It removes the mandate need to have a static config file defining your repos

*How the new flow looks like:*
* Typically it doesn't change anything in the flow except that now you don't necessarily need to create the config file for this to work.

*What's the use case for this?*
* For people with too large of dynamic repo set that they can't have a stick config file and hence want to track all the recent PRs for them/in their profile.

dlvhdr:
I'm working on removing the repo list requirement and think it will solve that for you. 
The approach you suggest will make it so new users start with prs from their notifications which is nit something I want. 
I would like to have a default config file. 
Once I allow not to pass the repos list, that will be easy to do.

dlvhdr:
See https://github.com/dlvhdr/gh-prs/pull/40

PR #86, opened 2022-02-16
Support multiple providers to open browser on linux

joaom00:
# Summary
This PR adds support for multiple providers on linux to open the browser and also for users using WSL. (I use WSL)

### Before

https://user-images.githubusercontent.com/48808846/154288664-3dfbc7c0-52eb-4b89-8669-45d1be59b8e0.mp4

### After

https://user-images.githubusercontent.com/48808846/154288730-ef3b97a9-9dd7-429e-bbdb-7783bc39367a.mp4

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -25,7 +25,7 @@ func Min(a, b int) int {
 	return b
 }
 
-func runProviders(url string) error {
+func openInBrowser(url string) error {
 	var err error
 	providers := []string{"xdg-open", "x-www-browser", "www-browser", "wslview"}
 
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


utils/utils.go
```diff
@@ -24,12 +25,29 @@ func Min(a, b int) int {
 	return b
 }
 
+func runProviders(url string) error {
```
dlvhdr, line 28: Thanks for the PR!
Looks good, but can you rename this method? Providers is a kind of general term.
```suggestion
func openInBrowser(url string) error {
```

dlvhdr (changes requested):

utils/utils.go
```diff
@@ -24,12 +25,29 @@ func Min(a, b int) int {
 	return b
 }
 
+func openInBrowser(url string) error {
```
dlvhdr, line 28: I actually didn't notice this function was just for linux, so please rename.
Also you forgot to change the call online 50
```suggestion
func openLinuxBrowser(url string) error {
```

dlvhdr (changes requested):

utils/utils.go
joaom00: Oh, sorry about that

dlvhdr (approved):

Issue #108, opened 2022-04-04
Some suggestions to the documents


wd:
I just want to address some issues I met and hope it could help others. 

## The folder that stores the config

According https://pkg.go.dev/os#UserConfigDir, the folder is different in each OS.  it is `~/Library/Application Support/gh-dash` on Mac. 

Don't waste your time struggling with the `~/.config/gh-dash` folder..

## The icons on the UI

You may not see the icons correctly even though you already have installed and used a nerd font. 

That is because some of the icons need to be rendered with the `bold` style. Make sure you have set the bold font correctly.

dlvhdr:
Fixed in #135

PR #133, opened 2022-05-27
Fix icons

mskelton:
A number of the icons for this extension were not showing up as shown in the screenshot.  I'm using the latest version of the Symbols Nerd Font which should work since it's a valid Nerd Font.

## Before

SCR-20220527-leq-2

## After

SCR-20220527-lgi

### DIFF
--- a/ui/components/issuesidebar/activity.go
+++ b/ui/components/issuesidebar/activity.go
@@ -57,7 +57,7 @@ func renderTitle() string {
 	return styles.MainTextStyle.Copy().
 		MarginBottom(1).
 		Underline(true).
-		Render(" Comments")
+		Render(" Comments")
 }
 
 func renderEmptyState() string {
--- a/ui/components/issuessection/issuessection.go
+++ b/ui/components/issuessection/issuessection.go
@@ -115,11 +115,11 @@ func (m *Model) UpdateProgramContext(ctx *context.ProgramContext) {
 func (m *Model) GetSectionColumns() []table.Column {
 	return []table.Column{
 		{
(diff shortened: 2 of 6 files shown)


dlvhdr:
Thanks for the contribution @mskelton.
Can you please first make sure you have a Bold version of your font installed? Might be related
See #108

mskelton:
So, I don't have a bold version of the font installed.  I use the [Kitty terminal](https://sw.kovidgoyal.net/kitty) which recommends the following approach for NERD fonts which doesn't use patch fonts, instead it uses a symbol map that maps to the NERD font symbols.

> If you are trying to use a font patched with [Nerd Fonts](https://nerdfonts.com/) symbols, don’t do that as patching destroys fonts. There is no need, simply install the standalone Symbols Nerd Font (the file NerdFontsSymbolsOnly.zip from the [Nerd Fonts releases page](https://github.com/ryanoasis/nerd-fonts/releases)). kitty should pick up symbols from it automatically, and you can tell it to do so explicitly in case it doesn’t with the [symbol_map](https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.symbol_map) directive:

Source: https://sw.kovidgoyal.net/kitty/faq/?highlight=nerd#kitty-is-not-able-to-use-my-favorite-font

I did try removing the bold styling from some of the broken icons and it still failed to render properly.

dlvhdr:
Does this work with other terminal emulators? I also use kitty. 
Anyway seems safe to me

dlvhdr (approved):

PR #134, opened 2022-05-27
Fix pending checks not always rendering as pending

mskelton:
# Summary

Pending checks that are defined in the `StatusContext` were not properly rendering as pending.  Check runs worked fine, but external CI services like Buildkite which use `StatusContext` were not rendering properly.

I also improved the formatting of the pending checks pill which was rendering strangely before.

Verified that pending checks work properly after this change.

### Before
image

### After

image

### DIFF
--- a/ui/components/prsidebar/prsidebar.go
+++ b/ui/components/prsidebar/prsidebar.go
@@ -94,7 +94,7 @@ func (m *Model) renderChecksPill() string {
 		return pillStyle.Copy().
 			Background(styles.DefaultTheme.WarningText).
 			Render(" Checks")
-	} else if status == "SUCCESS" {
+	} else if status == "PENDING" {
 		return pillStyle.Copy().
 			Background(styles.DefaultTheme.FaintText).
 			Foreground(styles.DefaultTheme.MainText).
--- a/ui/components/prsidebar/prsidebar.go
+++ b/ui/components/prsidebar/prsidebar.go
@@ -7,7 +7,6 @@ import (
 	"github.com/charmbracelet/lipgloss"
 	"github.com/dlvhdr/gh-dash/data"
 	"github.com/dlvhdr/gh-dash/ui/components/pr"
-	"github.com/dlvhdr/gh-dash/ui/constants"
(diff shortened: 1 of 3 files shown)


dlvhdr:
Looks good overall @mskelton!
I checked this locally and for me the checks are now incorrect.
At Wix, where I work, our CI probably reports things differently.

image

As you can see I have a green check, but the pill shows in gray.

mskelton:
Oops, forgot a hard coded conditional I changed for testing.  Fixed in https://github.com/dlvhdr/gh-dash/pull/134/commits/58a3563f1829482122812da3759bae874b5cddc6.

ui/components/prsidebar/prsidebar.go
```diff
@@ -98,7 +97,9 @@ func (m *Model) renderChecksPill() string {
 	} else if status == "PENDING" {
 		return pillStyle.Copy().
 			Background(styles.DefaultTheme.FaintText).
-			Render(constants.WaitingGlyph + " Checks")
+			Foreground(styles.DefaultTheme.MainText).
+			Faint(true).
+			Render(" Checks")
```
dlvhdr, line 102: Can you set the waiting glyph in the constants pkg? Instead of hardcoding it here

dlvhdr (changes requested):

ui/components/prsidebar/prsidebar.go
mskelton: The reason I hard coded it here was to fix it not displaying properly as I described in the PR description.  Before, it looked like this:

image

But hard coding it makes it display correctly:

image

mskelton: If you want, I could make constants for each of the icons used where each icon is just the actually glyph.  Then it could be referenced everywhere since it is just a string and not a lipgloss style.

dlvhdr: Ah ok, yeah the constants are with the styles already. Cool...
You could make the glyphs a constant but maybe that's just overkill...

dlvhdr (approved):

Issue #131, opened 2022-05-26
Support linux-386


celeronCoder:
**Is your feature request related to a problem? Please describe.**

When I try to install the extension with the following command:
```bash
gh extension install dlvhdr/gh-dash
```
I get the following error:
```error
gh-dash unsupported for linux-386. Open an issue: `gh issue create -R dlvhdr/gh-dash -t'Support linux-386'`
```

Here's my PC info:
```bash
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
Release:	20.04
Codename:	focal
```

On neofetch:
```log
           .-/+oossssoo+/-.               khushal@celeron 
        `:+ssssssssssssssssss+:`           --------------- 
      -+ssssssssssssssssssyyssss+-         OS: Ubuntu 20.04.4 LTS x86_64 
    .ossssssssssssssssssdMMMNysssso.       Kernel: 5.13.0-44-generic 
   /ssssssssssshdmmNNmmyNMMMMhssssss/      Uptime: 23 mins 
  +ssssssssshmydMMMMMMMNddddyssssssss+     Packages: 2076 (dpkg), 14 (snap) 
 /sssssssshNMMMyhhyyyyhmNMMMNhssssssss/    Shell: zsh 5.8 
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Resolution: 1920x1080 
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   DE: GNOME 
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   WM: Mutter 
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   WM Theme: Adwaita 
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   Theme: Yaru-dark [GTK2/3] 
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Icons: Yaru [GTK2/3] 
 /sssssssshNMMMyhhyyyyhdNMMMNhssssssss/    Terminal: gnome-terminal 
  +sssssssssdmydMMMMMMMMddddyssssssss+     CPU: Pentium E5700 (2) @ 2.991GHz 
   /ssssssssssshdmNNNNmyNMMMMhssssss/      GPU: Intel 82G33/G31 Express 
    .ossssssssssssssssssdMMMNysssso.       Memory: 2649MiB / 3916MiB 
      -+sssssssssssssssssyyyssss+-
[...]

Issue #113, opened 2022-04-10
Support FreeBSD


yonas:
I'd like to use gh-dash on FreeBSD.

PR #138, opened 2022-06-08
Update gh-extension-precompile

mislav:
- Use stable `@v1` version for this action. This ensures that gh-dash will continue to get fixes and other updates from this Action, such as new Go platforms added for binaries. The `v1` branch intends to never break backwards compatibility.
- Explicitly specify Go version used for building, otherwise it will be Go 1.16

Ref. https://github.com/cli/gh-extension-precompile/releases/tag/v1.2.0

Note that binaries produced by `gh-extension-precompile` follow a slightly different naming convention than `release.sh` in this repository: it uses `386` in the filename of `GOARCH=386` binaries instead of `i386`. When installing extensions, gh will look up available assets based on literal GOARCH, so `i386` will actually not match 32-bit systems.

Fixes https://github.com/dlvhdr/gh-dash/issues/131, fixes https://github.com/dlvhdr/gh-dash/issues/113

### DIFF
--- a/.github/workflows/go-release.yml
+++ b/.github/workflows/go-release.yml
@@ -13,4 +13,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: cli/gh-extension-precompile@v1.1.2
+      - uses: cli/gh-extension-precompile@v1
+        with:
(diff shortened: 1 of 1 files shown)


dlvhdr (approved):
Amazing, thank you! ❤️

Issue #129, opened 2022-05-19
[BUG] Subscribed Tab Doesn't Show Subscribed Repos/PRs


PatrickTulskie:
**Describe the bug**

The "subscribed" tab shows PRs for the gh-dash and cli repos. It doesn't show anything for repos that I am watching.

**To Reproduce**
1. Go to 'subscribed'
2. Observe a bunch of irrelevant entries.

**Expected behavior**

Should show PRs or repos that I am actually subscribed to.

**Screenshots**

NA

**Desktop (please complete the following information):**
 - Any platform

dlvhdr:
Yeah maybe the name is confusing, but it's just an example of a section you can create. Maybe I should look into showing actual repos you're subscribed to

PatrickTulskie:
Ohhh makes sense. I guess this isn't really a bug then. Maybe even just a readme update about that would be cool.

For some reason, my install doesn't have a config in the same location the readme expects it to be though. Should I just put one there and configure that tab that way? Or is there something else I should do?

dlvhdr:
Theres an issue about the config location. If you don't set the XDG variables I put it in a bad location. For now you can just put it under ~/.config/gh-dash/config.yml

PatrickTulskie:
Tried to put a config there and it didn't seem to get picked up so I went searching my drive. Found it in ~/Library/Application Support/gh-dash/config.yml

When I deleted that, it kept recreating the config there so for now I just symlinked it to ~/.config/gh-dash/config.yml and it seems to work now.

Until I opened this bug I didn't realize how configurable this was. Really awesome!

dlvhdr:
Yup that's the issue. If you don't have the XDG env variables set

dlvhdr:
This should be solved now with the latest version.

Issue #74, opened 2022-01-24
[BUG] Subscribed list is scoped to this repo and cli/cli


privatenumber:
**Describe the bug**
The "Subscribed" tab doesn't show anything relevant to me.

Looks like the Filter for it is scoped to `cli/cli` and this repository:
https://github.com/dlvhdr/gh-prs/blob/9f9aedad3bd392f990d96b722d4aff9fb893eaac/config/parser.go#L64

**Expected behavior**
To show issues/PRs that I'm subscribed to.

dlvhdr:
Hi, the subscribed tab is just provided as an example. 
You're free to change the config however you want.

micalevisk:
@privatenumber this is just the default config that will be used if you didn't have write your own [`config.yml`](https://github.com/dlvhdr/gh-prs#configuring). You should edit this file

privatenumber:
Gotcha. Thanks for explaining.

I guess the feedback is that it was a confusing experience then.

I would probably want more useful/relevant defaults but feel free to close this.

dlvhdr:
I'll try and think of better defaults

PR #156, opened 2022-08-17
Change default Subscribed tab to Involved

ceedubs:
# Summary

This changes the `Subscribed` tabs in the default config to `Involved`
tabs and changes the filter from hard-coded repos (such as `cli/cli`) to
`involves:@me`.

This addresses behavior that has surprised some folks such as in #74 and #129.

Ideally, we could fetch issues/PRs that the user has subscribed to, but
it looks like there's not an obvious way to do that. In its absence, the
`involves` filter might be a reasonable default that is less surprising
to people.

I changed my local config to match this and ran `gh dash` to observe the results.

Admittedly I did not set up a `go` environment to test these changes.

## Images/Videos

Screen Shot 2022-08-17 at 9 41 39 AM

### DIFF
--- a/config/parser.go
+++ b/config/parser.go
@@ -125,8 +125,8 @@ func (parser ConfigParser) getDefaultConfig() Config {
 				Filters: "is:open review-requested:@me",
 			},
 			{
-				Title:   "Subscribed",
-				Filters: "is:open -author:@me repo:cli/cli repo:dlvhdr/gh-dash",
+				Title:   "Involved",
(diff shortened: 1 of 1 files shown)


dlvhdr:
Yeah that's cool, thanks @ceedubs

dlvhdr (approved):

ceedubs:
I ended up also adding `archived:false` to most of my filters. That might be a reasonable thing to add to defaults as well, but I didn't want to make too many opinionated changes in the same PR.

PR #177, opened 2022-09-21
fix(ui): preview unrenewed when default view is `issues`

miles170:
# Summary

Change the default view to `issues`, then verify that the preview has been updated.


dlvhdr:
Thank you!

dlvhdr (approved):

PR #189, opened 2022-10-19
feat(pagination): fix #112 Implement pagination for issues section 

### LINKED ISSUE #112: Add a 'Load more items' functionality
With lots of issues and/or PRs, the default displayed items count might come short.

A key binding which would load more items in the current section.

Increasing the loaded items in the config file. But this makes dash slower to load, and eventually you'll hit the limit.

rambhosale:
This Pr adds the functionality for loading more issues when the cursor reaches the 5th last item in the issues

This is the inital working draft, if there are no issues with this PR and its implementation will complete the same feature for the PRs section

On reaching the 5th last issue fetch n more issues (based on issuesLimit value) 

## Images/Videos
default View
image

On reaching/ crossing the 5th last issue load more
image

### DIFF
--- a/ui/components/issuessection/issuessection.go
+++ b/ui/components/issuessection/issuessection.go
@@ -21,6 +21,7 @@ const SectionType = "issue"
 type Model struct {
 	section.Model
 	Issues []data.IssueData
+	Cursor data.PageInfo
 }
 
 func NewModel(id int, ctx *context.ProgramContext, cfg config.IssuesSectionConfig) Model {
@@ -35,6 +36,7 @@ func NewModel(id int, ctx *context.ProgramContext, cfg config.IssuesSectionConfi
 			"Issues",
 		),
 		[]data.IssueData{},
+		data.PageInfo{},
 	}
 
 	return m
@@ -58,10 +60,17 @@ func (m Model) Update(msg tea.Msg) (section.Section, tea.Cmd) {
 			case msg.Type == tea.KeyEnter:
 				m.SearchValue = m.SearchBar.Value()
 				m.SetIsSearching(false)
+
+				// reset the cursor position when fetching the for the new filter
+				m.Cursor = data.PageInfo{}
 				return &m, m.FetchSectionRows()
 			}
 
 			break
+			// when the cursor is on the 5th last item load more items
+			// TODO: Add functionality for customizing this number
+		} else if msg.Type == tea.KeyDown && m.Table.GetCurrItem()+5 == m.Table.GetLastItem() {
+			return &m, m.FetchSectionRows()
 		}
 
 		switch {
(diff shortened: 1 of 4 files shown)


dlvhdr:
This is awesome! Thank you! I'll take a look later this weekend

ui/components/issuessection/issuessection.go
```diff
@@ -58,10 +60,17 @@ func (m Model) Update(msg tea.Msg) (section.Section, tea.Cmd) {
...
+
+				// reset the cursor position when fetching the for the new filter
+				m.Cursor = data.PageInfo{}
 				return &m, m.FetchSectionRows()
 			}
 
 			break
+			// when the cursor is on the 5th last item load more items
+			// TODO: Add functionality for customizing this number
+		} else if msg.Type == tea.KeyDown && m.Table.GetCurrItem()+5 == m.Table.GetLastItem() {
```
dlvhdr, line 72: * Let's extract the 5 to a const for now

ui/components/issuessection/issuessection.go
```diff
@@ -222,10 +233,14 @@ func (m *Model) FetchSectionRows() tea.Cmd {
...
+
+	if !m.Cursor.HasNextPage {
+
+		m.Issues = nil
+		m.Table.ResetCurrItem()
+		m.Table.Rows = nil
+		m.IsLoading = true
+	}
 	var cmds []tea.Cmd
 	cmds = append(cmds, m.CreateNextTickCmd(spinner.Tick))
```
dlvhdr, line 245: I don't think we need this when we fetch the next rows, as this spinner is showing only on initial render, when we don't have any rows yet

ui/components/issuessection/issuessection.go
```diff
@@ -222,10 +233,14 @@ func (m *Model) FetchSectionRows() tea.Cmd {
 	if m == nil {
 		return nil
 	}
-	m.Issues = nil
-	m.Table.ResetCurrItem()
-	m.Table.Rows = nil
-	m.IsLoading = true
+
+	if !m.Cursor.HasNextPage {
```
dlvhdr, line 237: I think I added the reset logic here because of when the user refreshes the section. So please check that refresh still works

ui/components/issuessection/issuessection.go
```diff
@@ -234,20 +249,29 @@ func (m *Model) FetchSectionRows() tea.Cmd {
...
 
+		// append the newely fetched issues list to the already fetched list
+		// only when paginating request is request
+		if m.Issues != nil {
+			fetchedIssues = append(m.Issues, fetchedIssues...)
+		}
+
 		sort.Slice(fetchedIssues, func(i, j int) bool {
-			return fetchedIssues[i].UpdatedAt.After(fetchedIssues[j].UpdatedAt)
+			return fetchedIssues[i].Number > fetchedIssues[j].Number
```
dlvhdr, line 266: If you do this, please add the `sort:updated` search query by default

dlvhdr (changes requested):

ui/components/issuessection/issuessection.go
rambhosale: When ```m.IsLoading``` is set to ```true``` while fetching the new rows there is a flash behavior (list disappears and appears again), so I excluded setting it while fetching the next rows

Issue #221, opened 2023-01-16
[BUG] dash should not list issues of archived repositories


robdimsdale:
**Describe the bug**

Similar to #124 but for issues instead of PRs. Currently issues on archived repositories are shown, despite the fact that users generally can't interact with issues on archived repositories. Users cannot close issues, comment on them, etc.

I'm happy to submit a PR to address this if we agree we should filter out issues from archived repositories.

**To Reproduce**
1. run `gh dash` with some config which will fetch issues of an archived repository (e.g. https://github.com/paketo-buildpacks/stacks)
2. observe that archived issues are visible

**Expected behavior**
I expect that issues from archived repositories are not shown.

**Screenshots**

Here is an archived repo, showing open issues:

<img width="941" alt="Screen Shot 2023-01-16 at 3 49 10 PM" src="https://user-images.githubusercontent.com/7230694/212763476-0033d883-0477-4969-b832-8dd848bb05f4.png">

And just to demonstrate that this repository is archived:

<img width="478" alt="Screen Shot 2023-01-16 at 3 49 33 PM" src="https://user-images.githubusercontent.com/7230694/212763526-2d728acc-37de-4990-94fb-1405ae2f94ef.png">

**Desktop (please complete the following information):**
 - OS: Mac

dlvhdr:
Yes this is tech debt I want handled. Currently the code between prs and issues is a bit duplicated with the original thought to merge once needed. Since I didn't find the time, some features were only implemented in one place.
[Here's the code for prs.](https://github.com/dlvhdr/gh-dash/blob/0b1a3eaa5291f046b32d9e6a34c4e442c5f683d0/ui/components/prssection/prssection.go#L273)

robdimsdale:
@dlvhdr I think this is good to close now, yes? Unless you'd prefer to keep the issue open to drive out a refactor?

PR #223, opened 2023-01-17
fix: #221 filtered issues in archived repositories

yokishava:
# Summary
I've tried to filter archived pull requests before, but I saw this issue
We have done the same for issues in archived repositories.

#221

I also saw a comment that the code was becoming redundant. I considered modifying the code to take that into account, but decided against it because the scope of the change would be too large.

- remove `excludeArvhicedPullRequests` in prsection.go
- remove `sort` in prsection.go and issuesection.go
- filtered issues and sorted issues in FetchIssues
- filtered pull requests and sorted pull requests in FetchPullRequests
- remove test.yaml in github workflows

## Images/Videos

before
スクリーンショット 2023-01-17 19 14 08

after
スクリーンショット 2023-01-17 19 14 44

I used this repository to check if archived issues are filtered.
https://github.com/yokishava/element


robdimsdale:
This is almost exactly what I was planning on doing to address #221 - thank you for beating me to it, @yokishava !

I am curious to learn more about why you removed the test. I was planning on adding a test to `issuesection` (or `issueapi`) similar to what you previously added for PRs. But I see you removed it.

I think there's still value in testing the archived/sort functionality - do you agree ? If so, should we add the test back in?

yokishava:
Thank you for comment!

The reason for removing the tests is that the changes made it harder to write tests.

You are right that it would be better to have tests, but if we are going to write tests, I think we need to make more changes to FetchIssue and FetchPullRequests to get there. I have lowered the priority of the tests to reduce the scope of the changes.

If you still think it would be better to have tests, I would be happy to modify them too.

robdimsdale:
> The reason for removing the tests is that the changes made it harder to write tests.

Ah, yes. The `issueapi`/`prapi` is tightly coupled to the GitHub api, so in order to test the archiving and sort functionality, we'd either have to refactor the relevant functions to consume a mockable interface, or we'd have to commit to integration-testing against a real, archived, repository.

I agree that both of those seem out of scope for this change, so I agree with removing the tests.

dlvhdr:
I'm good with this as well.

data/issueapi.go
```diff
@@ -101,8 +99,16 @@ func FetchIssues(query string, limit int, pageInfo *PageInfo) (IssuesResponse, e
 
 	issues := make([]IssueData, 0, len(queryResult.Search.Nodes))
 	for _, node := range queryResult.Search.Nodes {
+		if node.Issue.Repository.IsArchived {
+			continue
+		}
 		issues = append(issues, node.Issue)
 	}
+
+	sort.Slice(issues, func(i, j int) bool {
```
dlvhdr, line 108: Can you please create an issue that removes the sorting code, and adds a default `sort:updated` like in https://github.com/dlvhdr/gh-dash/pull/223

dlvhdr (approved):

data/issueapi.go
robdimsdale: I'm not sure I follow the reference - this issue is `#223` and when I searched existing issues and PRs for `sort` I don't see any other issue that I could reference.

dlvhdr: I swear I pasted a link to the Github filters docs :D

robdimsdale: done: https://github.com/dlvhdr/gh-dash/issues/225

yokishava: Thanks for review!
I added sort qualifier to `makeIssuesQuery` and `makePullRequestsQuery` .

Issue #217, opened 2023-01-09
Make repo name truncation configurable


jlec:
The repo name is truncated at a hard 18 chars

https://github.com/dlvhdr/gh-dash/blob/d89be39ea4a8d930ee6f8e33eaa0cd982e758006/ui/components/pr/pr.go#LL143C64-L143C66

I do have a lot of repo starting with `ansible-..` or `terraform-...`. To properly understand which repo we are talking about I would need more chars. Could we make this configurable?

robdimsdale:
@dlvhdr Can I ask if there is value in continuing to support truncation on repository names. As someone who often interacts with repositories that have long names, I actually prefer to increase column width and disable truncation (by modifying the code and rebuilding). And I can't find an easy way to make this configurable (see details below).

I have to assume that most users who interact with repository names longer than the hard-coded limit of 18 characters would prefer to remove truncation altogether. And users with repository names less than the hard-coded limit are unaffected either way.

I am happy to submit a PR that removes truncation altogether, if that's the direction you are willing to go in.

### Configuring truncation length

If there is value in truncating the repo name, then I think it's actually quite tricky to make this configurable. Taking `PRs` as an example (though the same logic applies to `issues` too) - the truncation happens when creating the contents [here](https://github.com/dlvhdr/gh-dash/blob/0b1a3eaa5291f046b32d9e6a34c4e442c5f683d0/ui/components/pr/pr.go#L143) - before the `TextStyle` is applied and the cell is rendered. But there is currently no mechanism to provide the column width at this point. The column width is not calculated until the entire table is rendered [here](https://github.com/dlvhdr/gh-dash/blob/0b1a3eaa5291f046b32d9e6a34c4e442c5f683d0/ui/components/table/table.go#L218) - after the row is created and the `TextStyle` applied.

I explored truncating the text at this point that the table is rendered but it does not work as expected due to the text styling adding extra characters. And there is no obvious way to remove the text styling, truncate the contents, and re-add the text styling. The number of characters added by styling isn't consistent, so we can't just add an offset when truncating. And we can't "undo" the styling, because it's just a string.

### Commit history for truncation

It looks like truncation used to be present for other fields - specifically title and author. But truncation for those fields was removed in [this commit](https://github.com/dlvhdr/gh-dash/commit/b1ef660a8cbd267d175a041c80d0d04628f9e0a3) - leaving only truncation for repo name. It's not clear why truncation was removed for these other fields but kept for repo name.

Additional commit history:
[...]

PR #224, opened 2023-01-17
Remove truncation of repo name.

robdimsdale:
# Summary

Removes truncation on repo name and adds full name + number to the sidebar. See #217 .

Manually, with a repo name longer than the default of 18 characters.

### DIFF
--- a/ui/components/issue/issue.go
+++ b/ui/components/issue/issue.go
@@ -39,7 +39,7 @@ func (issue *Issue) renderUpdateAt() string {
 }
 
 func (issue *Issue) renderRepoName() string {
-	repoName := utils.TruncateString(issue.Data.Repository.Name, 18)
+	repoName := issue.Data.Repository.Name
 	return issue.getTextStyle().Render(repoName)
 }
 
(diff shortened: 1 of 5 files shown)


dlvhdr:
Looks good! Can you please provide a screenshot of how the sidebar looks with this change?

robdimsdale:
Sure - here's an example:

Screen Shot 2023-01-18 at 2 31 31 PM

robdimsdale:
Pushed a commit with that change. Here's a screenshot:

Screen Shot 2023-01-19 at 8 29 24 AM

robdimsdale:
Addressing [this question](https://github.com/dlvhdr/gh-dash/issues/217#issuecomment-1387621541) here:

> Cool, so without grow=true, how does that look? Can you add a screenshot?

The screenshots above actually do not have `grow=true` - I just increased the column width in the config.

Here is a smaller (default) column width - smaller than a repo name:

Screen Shot 2023-01-19 at 8 44 33 AM

And here is the same data with a manually-increased column width:

Screen Shot 2023-01-19 at 8 45 12 AM

dlvhdr:
Awesome, looks good!

PR #226, opened 2023-01-19
Add assignees to PRs.

robdimsdale:
# Summary

This PR adds the `assignees` field to PRs. It is essentially identical to what already exists for issues. The default configuration is for this field to be rendered (i.e. `hidden: false`) - as is the case for issues - and the length is `20` - again, the same as for issues.

I (and others on my team) use the assignment feature to keep track of PRs that we are responsible for, independent of who created the PR. It is helpful for us to see which PRs are unassigned or assigned to others.

I tested this change by configuring the dashboard to look at repositories with a mixture of assigned and unassigned PRs. See screenshot below.

## Images/Videos

Screen Shot 2023-01-19 at 1 02 58 PM


dlvhdr:
Awesome stuff, thank you!

Issue #228, opened 2023-01-19
Feature request: built-in support for assigning/unassigning issues and PRs


robdimsdale:
**Is your feature request related to a problem? Please describe.**

My workflow (and that of my team) leverages the ability to assign users/teams to issues and PRs. Currently, to achieve this from within `gh-dash`, I have to either create a custom command - which lacks the feedback and automatic refreshing of a built-in command - or I have to open the repo in a browser and assign a user from the UI. Neither mechanism is optimal.

**Describe the solution you'd like**

I would like `gh-dash` to have built-in support for assigning users/teams. I'm less concerned about unassigning, but I think for completeness and symmetry if we implement assignment, we should also implement unassignment. If nothing else it will make testing easier.

The minimal successful outcome would be to assign only to myself. That would actually solve for my workflow. However, I think it's probably a better user experience for the assignment (and unassignment) to allow user input.

I'd imagine that the assign/unassign text input would be a list of usernames, ~~comma-separated~~ whitespace-separated. I've validated that you can assign and unassign multiple users in a single `gh issue edit` invocation. The default text for assignment would be: `@me`. The default text for unassignment would be the list of currently assigned users.

**Describe alternatives you've considered**

At a high level, the only alternatives are described above - custom commands or opening the web browser and assigning from there.

In terms of the implementation details, I would be happy with hard-coding the assignment/unassignment to `@me` but as mentioned above, I think it's a much better user experience to provide text input with pre-populated values.

**Additional context**

I'm happy to implement this, once we have consensus on the UX.

robdimsdale:
Completed via #229

PR #229, opened 2023-01-20
Add assign/unassign functionality to issues and pull requests.

robdimsdale:
# Summary

This PR adds the ability to assign to and unassign from issues and PRs. It addresses most of the functionality in #228. This PR is intended to drive out the UX of this feature, and as such is likely to change while we determine the best UX.

Highlights:
* Assign one or more users to an issue/PR with `a`
* Unassign one or more users from an issue/PR with `A`
* Assigning/Unassigning gives a modal similar to commenting - with the same keys to submit and cancel (`ctrl+d` and `crtl+c`/`esc` respectively)
* Supports the same user syntax as `gh issue edit --add-assignee <user>` (e.g. `@me` assigns the current user, `--add-assignee` multiple times for multiple users).

There are probably opportunities to refactor the code a few different ways - I just wanted to get a proof of concept up so we can concretely talk about the UX.

The main issue I encountered was that i can't figure out how to pre-populate the text field. E.g. when adding an assignee it would be good to pre-populated with the currently logged-in user (or `@me`), and when unassigning it would be good to pre-populate with the existing list of users.

Assigning and unassigning against a repo that can support multiple assignees. See below video gif.

## Images/Videos

Here is a demo gif of assigning/unassigning a single user and multiple users to an issue. I didn't create a similar gif for PRs but it works identically.

demo


dlvhdr:
Yeah the main thing to think about the UX.
I like that you added the title of `Assign users (whitespace-separated).
Do we want to repurpose the comment text view for this? My main intuition is that this looks pretty good! It does seem to work for all the other `gh pr edit` and `gh issue edit` commands.
Let's go for it :)

robdimsdale:
Do you think the assign box should be the same height (i.e. number of lines) as the comment box? I don't think I mind either way, just assumed that users would typically require fewer lines to assign that to comment. But, I think it's fine either way, so we can repurpose the comment box for this use-case.

We would need to refactor the comment box to take a prompt (or a use-case) so it can display the appropriate prompt (e.g. "Leave a comment", "Assign users (whitespace-separated)", or "Unassign users (whitespace-separated)".

Other than that - do you have any other feedback you'd like to see addressed while I'm implementing this?

dlvhdr:
Yeah I think I'm good with this overall.
Maybe in addition to the prompt being configurable, we should also allow to pass in the number of rows. Then, fields that only require a single row, like the title of a PR can be more intuitively represented.

ui/components/assignbox/assignbox.go
```diff
@@ -0,0 +1,107 @@
...
+	textArea   textarea.Model
+	assignHelp help.Model
+}
+
+var assignKeys = []key.Binding{
+	key.NewBinding(key.WithKeys(tea.KeyCtrlD.String()), key.WithHelp("Ctrl+d", "submit")),
+	key.NewBinding(key.WithKeys(tea.KeyCtrlC.String(), tea.KeyEsc.String()), key.WithHelp("Ctrl+c/esc", "cancel")),
+}
+
+func NewModel(ctx *context.ProgramContext) Model {
```
dlvhdr, line 25: Let's repurpose the comment box for this (rename to input box maybe)

robdimsdale:
Ok, that makes sense.

I think the way I'd do that is instead of passing in a prompt (and number of lines) when the box is rendered, passing those parameters at instance creation time. We'd end up with `var assignBox, unassignBox, commentBox`, each with their own parameters.

Make sense? Any concerns?

ui/components/issuesidebar/issuesidebar.go
```diff
@@ -97,6 +151,14 @@ func (m Model) View() string {
 		s.WriteString(m.commentBox.View())
 	}
 
+	if m.isAssigning {
+		s.WriteString(m.assignBox.View(true))
```
dlvhdr, line 155: In bubbletea, state like this (whether we're assigning or unassining) is held on the model - in this case the `assignbox`'s model.
To update this value, you usually just do `m.assignBox.setAssignState(AssignBox.Assigning)` or something like that.
The `View` method then reads the model and renders accordingly

ui/ui.go
```diff
@@ -118,12 +118,36 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return m, cmd
 		}
 
+		if m.prSidebar.GetIsAssigning() {
```
dlvhdr, line 121: I think we can just merge with the condition above: `if m.prSidebar.GetIsAssigning() || m.prSidebar.GetisCommenting...`

ui/ui.go
```diff
@@ -317,11 +365,31 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.syncSidebar()
 	}
 
+	if m.prSidebar.GetIsAssigning() {
```
dlvhdr, line 368: same

dlvhdr:
> Ok, that makes sense.
> 
> I think the way I'd do that is instead of passing in a prompt (and number of lines) when the box is rendered, passing those parameters at instance creation time. We'd end up with `var assignBox, unassignBox, commentBox`, each with their own parameters.
> 
> Make sense? Any concerns?

Hmm maybe, I guess you can see what's better when implementing it.
My intuition is to have 1 `inputBox` model and then changing it's model by calling `inputBox.setNumberOfLines()` and `inputBox.setPrompt()` or something.
But, if that proves to be harder to read or maintain then having them separate also works for me.

robdimsdale:
> > Ok, that makes sense.
> > I think the way I'd do that is instead of passing in a prompt (and number of lines) when the box is rendered, passing those parameters at instance creation time. We'd end up with `var assignBox, unassignBox, commentBox`, each with their own parameters.
> > Make sense? Any concerns?
> 
> Hmm maybe, I guess you can see what's better when implementing it. My intuition is to have 1 `inputBox` model and then changing it's model by calling `inputBox.setNumberOfLines()` and `inputBox.setPrompt()` or something. But, if that proves to be harder to read or maintain then having them separate also works for me.

That works too. I don't have a good enough mental model of the way bubble tea handles the lifecycle of components - when you can set/should set state. If your suggestion works then I prefer that.

dlvhdr:
Yeah basically state should be updated in `Update()` by changing the models directly (`inputBox.setNumberOfLines()` etc.) and `View()` should be a function that given a model, outputs a string.

robdimsdale:
Alright, I updated the PR with your feedback. I think it's much cleaner now.

I think there are still opportunities for refactoring, but I'd like to hold off until we're happy with the architecture of this PR - how and where `inputbox` state is set, etc.

ui/components/issuessection/issuessection.go
```diff
@@ -89,6 +89,12 @@ func (m Model) Update(msg tea.Msg) (section.Section, tea.Cmd) {
 				if msg.NewComment != nil {
 					currIssue.Comments.Nodes = append(currIssue.Comments.Nodes, *msg.NewComment)
 				}
+				if msg.AddedAssignees != nil {
+					currIssue.Assignees.Nodes = append(currIssue.Assignees.Nodes, msg.AddedAssignees.Nodes...)
```
dlvhdr, line 93: Could this cause duplicates to appear if I try to add a user that is already assigned to the issue?

ui/ui.go
```diff
@@ -112,13 +112,19 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return m, cmd
 		}
 
-		if m.prSidebar.GetIsCommenting() {
+		if m.prSidebar.GetIsCommenting() ||
+			m.prSidebar.GetIsAssigning() ||
+			m.prSidebar.GetIsUnassigning() {
```
dlvhdr, line 117: Maybe we can have the `prSidebar` just expose a single method called `m.prSidebar.isTextInputFocused()` or something

dlvhdr:
Looks great @robdimsdale, feel free to make refactors to reuse more code between issues and PRs if you'd like. Other than that, let me know when you're ready to merge and I'll merge it.

ui/components/issuessection/issuessection.go
robdimsdale: oh, great catch!

yes, it does render duplicates - at least until you refresh and it loads the correct list from GitHub.

I'll fix this. Thanks for catching it.

ui/ui.go
robdimsdale: Yeah, this sort of thing was one of the refactors I was thinking of exploring.

robdimsdale:
Alright, I fixed up the bug about rendering already-assigned users again. I also did a couple of refactors and small clean-ups.

More significantly - I also took the opportunity to pre-populate the assignment/unassignment text boxes with default values. Specifically, assigning defaults to the current user (or empty if the current user is already assigned), and unassignment defaults to the currently-assigned users (or empty if there are no current assignees).

I think this significantly improves the UX, but let me know if you disagree.

I can also capture a video of this new assignment/unassignment behavior if you want that.

robdimsdale:
Also @dlvhdr I think I'm happy with refactors for now. If you're happy with this feel free to merge it in - thanks!

dlvhdr:
Looks amazing @robdimsdale :)
Just a small comment, can you make the assignees column hidden by default for PRs?
I think a more useful thing to have there is the reviewer column, but I would hide that as well by default.

PR #252, opened 2023-04-14
Wildcard matching builtin

### LINKED ISSUE #250: [BUG] RepoPath wildcard matching does not work for built-in commands
**Describe the bug**

Built-in commands (like "checkout") do not work with the wildcard matching introduced in #150 - this feature was only implemented for custom commands.

**To Reproduce**

1. Create a wild card `repoPath`
2. Try to checkout a PR locally with `C` that should match the `repoPath` wildcard
3. See error: `Local path to repo not specified, set one in your config.yml under repoPaths`

**Expected behavior**

I would expect that setting wildcard matches under `repoPaths` should work for built-in commands as well as for custom commands.

 - OS: [e.g. Mac]

** Suggested solution**

I think we can extract [`ui.getRepoLocalPath`](https://github.com/dlvhdr/gh-dash/blob/50cd7818576a43f50d7523d20abe0ddff62fb4b8/ui/modelUtils.go#L64) into a separate package and then reference it from both the custom commands and the built-in commands.

We have to extract it to a separate package in order to avoid an import loop between the `ui` and the `prsection` package.

robdimsdale:
# Summary

This PR adds support for using wildcard paths for repo lookup for built-in commands in addition to the existing support for custom commands. Currently only custom commands support wildcard lookups for repo paths.

I had to refactor the existing code into a third package to avoid an import loop.

Additionally, as part of this refactor, I was able to improve the error handling for custom commands when the repo lookup failed.

Finally, I added unit tests for the wildcard matching, as I felt it was the easiest way to track the behavior of this function. I used vanilla golang testing - I did not add any additional libraries.

Fixes #250 

I confirmed that this change allows wildcard matching for built-in commands via manual testing. I used the `C`/`checkout` command as that interacts with the file system. For improving the custom command error handling, I used the example from the README:

```
keybindings:
  prs:
    - key: v
      command: >
        cd {{.RepoPath}} &&
        code . &&
        gh pr checkout {{.PrNumber}}
```

For the built-in commands, before the change I would see the error:

```
Local path to repo not specified, set one in your config.yml under repoPaths
```

After the change I would see the command succeed.

For the custom commands, previously it would give an unhelpful error:

```
Whoops, got an error: exit status 127
```

And now it gives the more familiar (and useful) error:

```
Local path to repo not specified, set one in your config.yml under repoPaths
```

All tests were done with a valid config for wildcard paths.

## Images/Videos

I can add them if we think this is necessary.

### DIFF
--- a/ui/common/repopath.go
+++ b/ui/common/repopath.go
@@ -5,17 +5,29 @@ import (
 	"strings"
 )
 
-// support [user|org]/* matching for repositories
-// and local path mapping to [partial path prefix]/*
-// prioritize full repo mapping if it exists
-func GetRepoLocalPath(repoName string, cfgPaths map[string]string) string {
+// GetRepoLocalPath returns the local path for a given repo name.
+// It will return the path if it exists in the config, or if the
+// repo name matches a wildcard path in the config.
+// It will prioritize exact repo name matches over wildcard matches.
+// If the second return value is true, the first return value is guaranteed
+// to be a valid path.
+// If the second return value is false, the first return value is undefined.
+// For a given config of:
+//
+//	{
+//	  "user/repo": "/path/to/user/repo",
+//	  "user_2/*":  "/path/to/user_2/*",
+//	}
+//
+// GetRepoLocalPath("user/repo", config) will return: "/path/to/user/repo", true
+// GetRepoLocalPath("user_2/some_repo", config) will return: "/path/to/user_2/some_repo", true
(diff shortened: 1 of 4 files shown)


dlvhdr:
Looks great! Thanks for this :)

dlvhdr (approved):

robdimsdale:
@dlvhdr can we get a new release with this bug fix please?

PR #262, opened 2023-04-30
chore: update gh-go to v2

### LINKED ISSUE #243: "Add support for secure token storage"
### Token storage change in [latest release](https://github.com/cli/cli/releases/latest) of [gh](https://github.com/cli/cli)

This is a message from the GitHub CLI team, maintainers of `gh`, writing to inform you that the most recent release of `gh` contains changes which may affect your extension. The latest release introduces the feature of storing authentication tokens in the system keyring (encrypted storage) instead of in a plain text file.
The keyrings that are supported are:

- Keychain on macOS

- GNOME Keyring on Linux (Secret Service dbus interface)

- Wincred on Windows

This has huge security benefits for the users of our tool and was one of our oldest outstanding [issues](https://github.com/cli/cli/issues/449). Unfortunately this change has the potential to break extensions that rely on utilizing the users authentication token to work.

In order to have continued compatibility with `gh` there are some actions you, as an extension author, need to take. These actions will depend on the implementation of your extension.

### Extensions built in Go using [go-gh](https://github.com/cli/go-gh):

1. Upgrade your `go-gh` version to v1.2.1, the latest version.

    - This can be done using `go get github.com/cli/go-gh@v1.2.1`

2. Verify that in your extension retrieval of the user authentication token is done using the  `auth.TokenForHost` function.

    - If you were previously accessing the authentication token using any other method it will no longer work.
    - Automatic resolution of the authentication token when using the API clients will continue to work without changes.

### All other extensions:

1. Verify that in your extension retrieval of the user authentication token is done by shelling out to the `gh auth token` command.

    - If you were previously accessing the authentication token using the `gh config get` command, reading the configuration file directly, or any other methods it will no longer work.

As of right now storing the authentication token in the system keyring is an opt-in feature, but in the near future it will be required and at that point if the changes above are not made then your extension will be broken for all users. If you have any questions/concerns about this change please feel free to open a [discussion in the gh](https://github.com/cli/cli/discussions) repo.

Thanks,
The GitHub CLI Team

wwsean08:
# Summary
Fixes #243 by updating go-gh from v0.1.0 to v2.0.0 (the latest release) as well as future proofs against breaking changes given that they are unlikely to make more breaking changes for a while.

My config currently has no token as it's stored securely so it was unable to connect with the released version of gh-dash, so after making this change I validated that I was able to use gh-dash to query my open issues/prs as shown in the recordings below.

### With secure token before updating the go-gh library

https://user-images.githubusercontent.com/839261/235377798-d45a7813-6f03-4ea7-958a-b1a7900a7caa.mov

### With secure token update go-gh library

https://user-images.githubusercontent.com/839261/235377858-a548f05a-0d68-4d6f-9a5b-236d913331eb.mov


dlvhdr:
Thank you!

PR #280, opened 2023-05-31
add `$GH_DASH_CONFIG` environmental variable

razvanazamfirei:
# Summary

This PR adds functionality to specify a location for the configuration file using an environmental variable (`$GH_DASH_CONFIG`). Users can specify a default path for the config file outside of `$XDG_CONFIG_HOME/gh-dash/config.yml`.

Additionally, it refactors getExistingConfigFile and getDefaultConfigFileOrCreateIfMissing, and updates the documentation. 

* reads and generates configuration files in the default location when `$GH_DASH_CONFIG` is not set.
* reads and generates configuration files in location specified by `$GH_DASH_CONFIG` if `$GH_DASH_CONFIG` is set. If the directory does not exist, it creates it.
* accepts `--config` overrides.

## Images/Videos

N/A

### DIFF
--- a/config/parser.go
+++ b/config/parser.go
@@ -319,8 +319,6 @@ func (parser ConfigParser) getExistingConfigFile() (*string, error) {
 		os.Getenv("GH_DASH_CONFIG"), // If GH_DASH_CONFIG is empty, the os.Stat call fails
 		filepath.Join(xdgConfigDir, DashDir, ConfigYmlFileName),
 		filepath.Join(xdgConfigDir, DashDir, ConfigYamlFileName),
-		filepath.Join(userConfigDir, DashDir, ConfigYmlFileName),
-		filepath.Join(userConfigDir, DashDir, ConfigYamlFileName),
 	}
 
 	// Check if each config file exists, return the first one that does
--- a/config/parser.go
+++ b/config/parser.go
@@ -300,8 +300,6 @@ func (parser ConfigParser) createConfigFileIfMissing(configFilePath string) erro
 }
 
 func (parser ConfigParser) getExistingConfigFile() (*string, error) {
-	var err error
-	var dashConfigFile string
 	homeDir, err := os.UserHomeDir()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


config/parser.go
```diff
@@ -312,63 +310,62 @@ func (parser ConfigParser) getExistingConfigFile() (*string, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	dashConfigFile = filepath.Join(userConfigDir, DashDir, ConfigYmlFileName)
-	if _, err := os.Stat(dashConfigFile); err == nil {
-		return &dashConfigFile, nil
+	configPaths := []string{
+		os.Getenv("GH_DASH_CONFIG"), // If GH_DASH_CONFIG is empty, the os.Stat call fails
+		filepath.Join(xdgConfigDir, DashDir, ConfigYmlFileName),
```
dlvhdr, line 320: this appears twice

dlvhdr:
@michaeltlombardi seems like this PR fails the hugo build but I'm not sure what the error means. Any chance you can help out here?

dlvhdr (commented):
Looks good, just see my 1 comment

michaeltlombardi:
> @michaeltlombardi seems like this PR fails the hugo build but I'm not sure what the error means. Any chance you can help out here?

I'll dig in today or tonight and put up a fix.

config/parser.go
razvanazamfirei: Updated, thank you!

michaeltlombardi:
@dlvhdr, @razvanazamfirei - apologies for the delay, I have a docs-site fix up in #282 that resolves the build issues. Entirely my fault, in a recent update to the theme I rewrote the TOC building and didn't account for pages with a single heading. Quick fix, all updated.

I initially planned to just add the fix commit to this PR, but didn't have access, so I put up a separate PR. Once that lands, this PR can be rebased and the docs site should build.

dlvhdr:
Thanks! Just landed your pr

dlvhdr (approved):

PR #288, opened 2023-06-28
fix: colors when TERM=wezterm

caarlos0:
# Summary

Current version defaults to ASCII when `TERM=wezterm`, this fixes it.

```
go run .
```


dlvhdr:
Thank you!

PR #327, opened 2024-01-28
feat: add support for configuing a default local repository path

markfeinstein:
# Summary

Add fallback logic to the `GetRepoLocalPath` to allow a `default_path` to be configured to handle repository paths that aren't explicitly configured. 

Additional changes:
- Update deps
- Add testify module
- Refactor repo path unit test to use table tests

## How did you test this change?

- Ran unit tests
- Local testing with a repo that doeos not have a matching or wildcard definition in gh-dash
    - Perform a checkout of a pr
    - Open a pr in Octo.nvim

PR #344, opened 2024-03-04
Deprecate `default_path` support in `repoPaths` configuration

rmacklin:
# Summary

Now that we support the more general `:owner/:repo` template configuration (https://github.com/dlvhdr/gh-dash/pull/343), the prior `default_path` functionality (https://github.com/dlvhdr/gh-dash/pull/327) is redundant - e.g. this configuration:
```yml
default_path: ~/code/repos
```
can be achieved equivalently with:
```yml
:owner/:repo: ~/code/repos/:repo
```

Thus, we are deprecating the `default_path` syntax. It's still supported but will be removed in the next major version.

This is just a documentation change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ defaults:
     width: 60 # width in columns
   refetchIntervalMinutes: 30 # will refetch all sections every 30 minutes
 repoPaths: # configure where to locate repos when checking out PRs
-  default_path: ~/code/repos # fallback value if none of the other paths match
+  default_path: ~/code/repos # (deprecated) fallback parent directory if none of the other paths match
   :owner/:repo: ~/src/github.com/:owner/:repo # template if you always clone github repos in a consistent location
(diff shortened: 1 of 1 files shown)


dlvhdr:
Might be worth running a quick code search to see how many use this. Something like: `path:gh-dash/config.yml default_path`

rmacklin:
There are [2 results (and one commented-out value)](https://github.com/dlvhdr/gh-dash/assets/1863540/a0fd19e3-dc5a-4096-949c-a03fb5b8a2ed) for that query:
https://github.com/search?q=path%3Agh-dash%2Fconfig.yml%20default_path&type=code

That said, I'm not sure how representative this is, given that many folks might not push their dash config to a public GItHub repo.

dlvhdr:
Ok cool, then given this not very scientific way, lets merge this 😀

rmacklin:
Ok sounds good! I've opened a new PR to remove `default_path` instead: https://github.com/dlvhdr/gh-dash/pull/345

PR #352, opened 2024-04-04
Add date format option to defaults

### LINKED ISSUE #67: Allow configuring date format

when i look at a PR from December it is displayed as "1y ago"

i'd like to be able to configure a custom date format in the config.toml file

using iso8601 dates instead of relative dates across the board, also seems reasonable, but i imagine some folks like relative timestamps

**Additional context**

image

miniscruff:
Currently only affects the updated at field for PRs and issues.
Hoping to get feedback on this idea and help contribute to documentation or schema update as required.

Closes #67

# Summary

This config will use a simple yyyy-mm-dd format with a larger column layout. Not entirely sure why its 12 width and not 10, maybe there is some padding by default.
```yaml
defaults:
  layout:
    prs:
      updatedAt:
        width: 12
        grow: true
  dateFormat: "2006-01-02"
```

The special value of RELATIVE, or no value at all, will keep the existing relative dates.
```yaml
defaults:
  dateFormat: "relative"
```

## Images/Videos
gh-dash-updated-at-date-format

With format `2006-01-02`

### DIFF
--- a/config/parser.go
+++ b/config/parser.go
@@ -98,6 +98,7 @@ type Defaults struct {
 	View                   ViewType      `yaml:"view"`
 	Layout                 LayoutConfig  `yaml:"layout,omitempty"`
 	RefetchIntervalMinutes int           `yaml:"refetchIntervalMinutes,omitempty"`
+	DateFormat             string        `yaml:"dateFormat,omitempty"`
 }
 
 type Keybinding struct {
--- a/docs/data/schemas/defaults.yaml
+++ b/docs/data/schemas/defaults.yaml
@@ -15,6 +15,7 @@ schematize:
       - Only fetch 20 PRs and issues at a time for each section.
       - Display the PRs view when the dashboard loads.
       - Refetch PRs and issues for each section every 30 minutes.
+      - Display dates using relative values.
 
       For more details on the default layouts, see the documentation for [sref:PR] and [sref:issue]
       layout definitions.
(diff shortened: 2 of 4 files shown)


dlvhdr:
Looks simple enough. nit Can we have the value be lower case relative?

miniscruff:
> Looks simple enough. nit Can we have the value be lower case relative? Also, what about the time? HH:MM I think that since this is written in Go, we should just follow Go's time.Format: [go.dev/src/time/format.go](https://go.dev/src/time/format.go) and instruct users that's the format we expect

Correct, it is using the go time format, so adding some sort of docs link would probably be helpful. 

By lower case relative do you mean compare to `relative` or make it case insensitive by allowing any form of `Relative`?

dlvhdr:
Yeah lowercase relative is my preference. We do have a docs site that you can add to. It's a kind of hugo site.

miniscruff:
> Yeah lowercase relative is my preference. We do have a docs site that you can add to. It's a kind of hugo site.

Gotcha, on it. Do you by chance have any tips on getting hugo to run locally. I did some research and have used hugo before but can't seem to get it working.

dlvhdr:
Can you please attach a screenshot of how this looks with a custom date?

miniscruff:
> Can you please attach a screenshot of how this looks with a custom date?

Added

dlvhdr (approved):

PR #356, opened 2024-04-14
support modifying default keybindings prototype

### LINKED ISSUE #214: Remap UI keybindings
It's a bit of a usability annoyance. I'm used to using Vim so just using the home row is practical. Right now I need to do hand gymnastics to use `Ctrl+d` and `Ctrl+u` for navigating the preview bar and find it rather uncomfortable.

I'd like to be able to remap those keys in my configuration file so `J` and `K` (mind the case) are used instead.
Note I'm not 100% sure this isn't currently possible, but going through the README and the source code I couldn't find anything useful.

There aren't many. Essentially it's either hardcoding my preferences (not very considerate to others used to the current workflow) or making it configurable.
I'd say the current keybinding section may benefit from some UI action syntax or maybe a new section could be added for the UI interaction.

miniscruff:
Fixes #214

Open questions:
Not sure if this mega switch case is really the way to go, I thought of creating an enum but either way we have a separate string to keybind map.

Not sure the yaml value for multiple words should be "switchview", "switchView" or some other method.

I had to change the Keys to a pointer so I can actually update them after the initial load. I am not sure if this is ok or if there is some other refactoring/change that would make more sense. For example, I tried putting a KeyMap on the shared ctx that has the config on it. And then just loading it on init, but it pushed more changes that I thought would be appropriate.

I am not sure how to handle the multiple key options for alts. I created a `Keys []string` that can be used that mimics `Key string` but with multiple. So it should be backwards compatible but not sure if there was another method you were expecting.

There is an error log on an invalid config, which will need to expand to an invalid config or after that, an invalid keybinding, such as a bad builtin. I am not sure I handled this properly, or if there is a better way, let me know.

# Summary
Allow modifying the builtin command keybindings.

I have a test config with the new universal option. Currently this is all I am testing.
```yaml
keybindings:
  universal:
    - key: v
      builtin: switchview
```

## Images/Videos
https://www.twitch.tv/videos/2176382326


ui/ui.go
```diff
@@ -33,7 +33,7 @@ import (
 )
 
 type Model struct {
-	keys          keys.KeyMap
+	keys          *keys.KeyMap
```
miniscruff, line 36: I am actually not sure this is used, it should be a duplicate of the pre-created KeyMap from the keys.go file.

config/parser.go
```diff
@@ -102,13 +103,35 @@ type Defaults struct {
...
+	Builtin string   `yaml:"builtin"`
+}
+
+func (kb Keybinding) NewBinding(previous *key.Binding) key.Binding {
+	helpDesc := ""
+	if previous != nil {
+		helpDesc = previous.Help().Desc
+	}
+
+	if len(kb.Keys) > 0 {
```
dlvhdr, line 118: Do you expect people to want alts? Is this a personal need of yours? I don't expect many users will actually define multiple keys for the same action.
If there's not a big use case, maybe it's better to omit to reduce the config bloat

ui/keys/keys.go
```diff
@@ -158,3 +159,52 @@ var Keys = KeyMap{
 		key.WithHelp("q", "quit"),
 	),
 }
+
+// Rebind will update our saved keybindings from configuration values.
+func Rebind(universal, issueKeys, prKeys []config.Keybinding) error {
```
dlvhdr, line 164: do you mean to handle `issueKeys` in a later PR?

ui/keys/keys.go
```diff
@@ -158,3 +159,52 @@ var Keys = KeyMap{
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func rebindUniversal(universal []config.Keybinding) error {
+	for _, kb := range universal {
+		log.Debug("Rebinding key", "builtin", kb.Builtin, "key", kb.Key)
```
dlvhdr, line 180: nice :)

ui/keys/keys.go
```diff
@@ -158,3 +159,52 @@ var Keys = KeyMap{
...
+		switch kb.Builtin {
+		case "up":
```
dlvhdr, line 182: I *think* you can use the keys.Keys global var here instead of the switch case. 
If you don't like updating a global var, you can expose a method in the `keys` package. I think this will be automatically picked up in the ui.go file

ui/keys/keys.go
```diff
@@ -158,3 +159,52 @@ var Keys = KeyMap{
...
+			Keys.Up = kb.NewBinding(&Keys.Up)
+		case "down":
+			Keys.Down = kb.NewBinding(&Keys.Down)
+		case "firstline": // should this be first-line or firstLine?
```
dlvhdr, line 186: I think `firstLine` would be more readable

ui/keys/prKeys.go
```diff
@@ -74,3 +75,45 @@ func PRFullHelp() []key.Binding {
 		PRKeys.WatchChecks,
 	}
 }
+
+func rebindPRKeys(keys []config.Keybinding) error {
```
dlvhdr, line 79: I see there's the `SetKeys` method on the KeyBinding so maybe you can update the global `Keys` struct that way?

dlvhdr (commented):
Looks like a great direction. I have some suggestions you play around with that might simplify the code

config/parser.go
miniscruff: Yea I wasn't entirely sure, I can't see myself particularly using it. So that is fair.

ui/keys/keys.go
miniscruff: It matches the pr keys stuff so I just didn't handle it yet with the questions I had, but it will be part of this PR.

ui/keys/keys.go
miniscruff: The switch is used to map the yaml string to the keybinding in the global var. I am not sure how to go bind the two otherwise. Other then maybe reflection.

config/parser.go
miniscruff: Removed in latest commit

ui/keys/keys.go
miniscruff: Using camel case for keys in next commit.

ui/keys/keys.go
miniscruff: Will be adding to prs and issue keys as well.

ws141:
Any chance for a new of this by @

dlvhdr:
@miniscruff would you like to resolve the merge conflict and merge?

miniscruff:
If I remember correctly I was running into issues still around keybinds used in both views not behaving properly. I can put in some time in the next few days to try and finish it.

ui/ui.go
```diff
@@ -279,6 +291,9 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 			return m, cmd
 
+            // I do not think this state machine works quite well enough with custom keybinds
+            // as it kinda breaks down if you use two different keys across the views.
+            // Is it worth refactoring this to first check the "view" before the key?
```
miniscruff, line 296: @dlvhdr just a question here about the keybind checks. I noticed we are checking for both PR and issue keys at once which sort of breaks if you bind to two separate keys. Wanted to clarify before I made any major changes.

dlvhdr: Yeah you're right, we would have to first check the current view

dlvhdr (dismissed):

ui/ui.go
miniscruff: Do you want me to make that change now or leave it for you?

dlvhdr: Making the change would be very appreciated 🔥

miniscruff:
Today is the day I finish, and now that I said it, it must come true.

Omnikron13:
At risk of causing any late-stage headaches here, I have to question if it might not be best to specify the view/context the other way around in the config format; have a keybinding specify which context or_ contexts_ it should apply in.

I don't know the full scope you (currently) want the project to cover, but there are theoretically quite a lot of different screens and contexts that could be added in future, which could result in some technical debt with this design. It could easily be desirable to have bindings that apply to multiple views, or to contexts which aren't strictly views but, well, broad contexts (dialogue boxes, text entry fields, lists/tables... even the current 'universal' tbh, though that option wouldn't be without it's own potential technical debt), and having the binding list where it applies is IMO a lot more robust, flexible, and future-proof (added bonus that you could disable a binding by commenting out it's context in the config file).

dlvhdr:
testing

dlvhdr:
Thanks for contributing this @miniscruff, it's very useful!

Issue #294, opened 2023-07-14
Duplicated records when refreshing several times.


orlandorode97:
**Describe the bug**
When the PR or Issue section is displayed the `refresh` or `refresh all` key is pressed several times causing duplicated records for both sections.

**To Reproduce**
1. Run `gh dash`.
2. Go to PR or Issue view.
3. Press `r` or `R` key for refreshing records several times.
4. The table of PRs or Issues is displayed with duplicated records.

**Expected behavior**
The table for PRs or issues should display consistent and accurate data without any duplicate records. Each row in the table should represent a unique entity or item.

**Screenshots**

https://github.com/dlvhdr/gh-dash/assets/34588445/27ce540c-cf77-4a11-b101-12fc1b2eb3af

**Desktop (please complete the following information):**
 - OS: **Mac OS Ventura 13.4**
 - Golang: **go version go1.20.3 darwin/arm64**
 - gh-dash: **gh dash  dlvhdr/gh-dash  v3.9.0**

PR #360, opened 2024-05-05
fix: duplicate record display on rapid consecutive refreshes

szktkfm:
# Summary
This PR addresses https://github.com/dlvhdr/gh-dash/issues/294

I have implemented a modification to retain the TaskID for each fetch task within the model. This TaskID is identical to the one included in the msg sent upon the completion of a record fetch. Saving taskID within the model is a Last-Write-Wins basis, ensuring that only the most recent TaskID is stored in the model.

The PR and issue lists within the model are updated only if the TaskID in the model matches the TaskID in the incoming message. 

I manually tested this change following the procedures shown in the video below. I confirmed that the issue was resolved without negatively impacting other functionalities such as paging or searching.

## Images/Videos

https://github.com/dlvhdr/gh-dash/assets/27483768/a78eb21e-f217-4771-8865-d3f2a76a32d6

### DIFF
--- a/ui/components/section/section.go
+++ b/ui/components/section/section.go
@@ -37,6 +37,7 @@ type Model struct {
 	PromptConfirmationBox     prompt.Model
 	IsPromptConfirmationShown bool
 	PromptConfirmationAction  string
+	LastTaskId                string
 }
 
 func NewModel(
--- a/ui/components/section/section.go
+++ b/ui/components/section/section.go
@@ -37,7 +37,7 @@ type Model struct {
 	PromptConfirmationBox     prompt.Model
 	IsPromptConfirmationShown bool
 	PromptConfirmationAction  string
-	LastTaskId                string
+	LastFetchTaskId           string
 }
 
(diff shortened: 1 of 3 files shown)


ui/components/section/section.go
```diff
@@ -37,6 +37,7 @@ type Model struct {
 	PromptConfirmationBox     prompt.Model
 	IsPromptConfirmationShown bool
 	PromptConfirmationAction  string
+	LastTaskId                string
```
dlvhdr, line 40: Can we rename this to `LastFetchTaskId` to make it clearer?

dlvhdr (commented):
Thank you! This is great!
Sorry for the delay :/ Just had a minor comment

ui/components/section/section.go
szktkfm: @dlvhdr 
Thank you for the suggestion. I've fixed.

dlvhdr (approved):

PR #362, opened 2024-05-19
Streamline custom commands

Omnikron13:
# Summary
Moved common functionality from `runCustomPRCommand()` & `runCustomIssueCommand()` into new `runCustomCommand()` function

Set and ran rudimentary custom commands in each context. Could probably do with some more robust testing though tbh.

I've left an empty `input` map still initialised in the new function, and merged in the specific data from the calls from different contexts, as I'm not sure what common data might want to be made available common to different contexts.

### DIFF
--- a/ui/modelUtils.go
+++ b/ui/modelUtils.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"maps"
 	"os"
 	"os/exec"
 	"text/template"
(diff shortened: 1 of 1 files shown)


Omnikron13:
Following the calls back, could drop `runCustomPRCommand()` & `runCustomIssueCommand()` too I suppose, unless there's likely to be extra logic there that doesn't want to be polluting `executeKeybinding()`.

dlvhdr (approved):

PR #366, opened 2024-05-23
Use "Switch to issues" and "Switch to PRs" for better observability

znd4:
# Summary

It was hard for me to find the capability to view issues -- I saw it mentioned in the README, but I couldn't find it in the `?` menu. This PR proposes varying the help text for the "switch view" action so that the words "issues" or "PRs" are present where applicable.

E.g.:

```
...
s Switch to issues
```

or

```
s Switch to PRs
```

Alternatively, we could use the simpler

```
s View issues
```

but that would lose the nice acronym. Also, "switch" feels like a relevant hint that this will fully replace the current view.

```sh
go run .
```

### DIFF
--- a/ui/keys/issueKeys.go
+++ b/ui/keys/issueKeys.go
@@ -8,6 +8,7 @@ type IssueKeyMap struct {
 	Comment  key.Binding
 	Close    key.Binding
 	Reopen   key.Binding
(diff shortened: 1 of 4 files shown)


dlvhdr:
Thank you @znd4!

dlvhdr (approved):

PR #369, opened 2024-05-24
Add labels to PR sidebar.

robdimsdale:
# Summary

This PR adds support for labels in the PR sidebar, essentially identical to how they already render for issues.

I leverage PR labels a lot in my workflow, so this is helpful to quickly see which PRs have (or are missing) labels.

## How did you test this change?

Add labels to PRs - see that they render similar to labels 

## Images/Videos

Here is an example PR with a label (PR can be found [here](https://github.com/robdimsdale/gh-dash/pull/1)).

Screenshot 2024-05-24 at 11 17 19

PR #378, opened 2024-05-28
Render labels cleanly across multiple lines.

robdimsdale:
# Summary

Fixes an issue where labels did not wrap neatly when their total length was longer than one line. (see https://github.com/dlvhdr/gh-dash/pull/369#issuecomment-2132224396). The previous behavior was inconsistent - sometimes it would wrap the label color but not the label text; sometimes it would fail to render labels at all.

I also backported the same logic to the issue sidebar, as it was already present there but we just hadn't discovered it yet.

With a PR (and issue) that had multiple labels associated with them

## Images/Videos

This is the new behavior. Notice the clean wrapping across multiple lines

Screenshot 2024-05-28 at 15 00 07

## Notes

The code is fairly clunky as it is manually building the multiple rows that consist of multiple lables. I couldn't find a native way to do this in the UI library (`lipgloss`). If there's a cleaner way to achieve this outcome I would prefer that!

### DIFF
--- a/ui/common/labels.go
+++ b/ui/common/labels.go
@@ -1,0 +1,45 @@
+package common
+
+import (
+	"github.com/charmbracelet/lipgloss"
+	"github.com/dlvhdr/gh-dash/data"
+)
+
+func RenderLabels(sidebarWidth int, labels []data.Label, pillStyle lipgloss.Style) string {
+	width := sidebarWidth
+
+	renderedRows := []string{}
+
+	rowContentsWidth := 0
+	currentRowLabels := []string{}
+
+	for _, l := range labels {
+		currentLabel := pillStyle.Copy().
+			Background(lipgloss.Color("#" + l.Color)).
+			Render(l.Name)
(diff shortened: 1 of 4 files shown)


dlvhdr:
thanks @robdimsdale, that's exactly how I would have done it, not sure there's a "better" way.
I would just avoid the code duplication and extract it to a common function.

robdimsdale:
Agreed @dlvhdr. I didn't want to tackle the refactor before being sure it was worth it. Specifically, if there was a one-liner way to achieve this in `lipgloss` then refactoring into a common package wouldn't be worth it.

Omnikron13:
Can't help but imagine the idea of dealing with TTY control codes, full & half width characters, zero width characters, combining characters, breaking & non-breaking characters, etc. etc. may have been off-putting to the authors of the native code? Could spelunk it I suppose and see if they either _say_ why, or at the very least don't do things in confusing ways eh.

PR #379, opened 2024-05-29
feat: make column titles configurable

alexkalderimis:
Prior to this change, column titles were hard-coded.

This change allows column titles to be defined in configuration.

# Summary

The current behavior requires the use of a nerd-font, which the user may not have or want. This allows titles to be configured manually, which allows for natural language titles.

Manual testing, using the following config section:

```yaml
prSections:
  - title: Mine
    filters: is:open author:@me
    layout:
      updatedAt:
        title: Age
      repo:
        title: Repo
      author:
        hidden: true
      lines:
        title: Lines
      ci:
        title: Checks
      reviewStatus:
        hidden: true
      state:
        hidden: true
```

## Images/Videos

Screenshot 2024-05-29 at 4 30 03 PM

### DIFF
--- a/config/parser.go
+++ b/config/parser.go
@@ -58,8 +58,9 @@ type PreviewConfig struct {
 }
 
 type ColumnConfig struct {
-	Width  *int  `yaml:"width,omitempty"  validate:"omitempty,gt=0"`
-	Hidden *bool `yaml:"hidden,omitempty"`
+	Width  *int    `yaml:"width,omitempty"  validate:"omitempty,gt=0"`
+	Hidden *bool   `yaml:"hidden,omitempty"`
+	Title  *string `yaml:"title,omitempty"`
 }
 
 type PrsLayoutConfig struct {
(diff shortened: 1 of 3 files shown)


dlvhdr (commented):
Thanks for the PR @alexkalderimis!

The PR looks clean but I think that if we want to support the lack of nerd fonts, it should probably be done in a more system wide level.
Not sure people would actually want to modify their column names and rather would just like to replace the nerd icons with something they can see/read.
So I was thinking of a simple `useNerdFonts: false` in the config.
We can then read that in multiple places and fall back to text if needed.

Another spot we can replace icons with text is in the `ui/constants/constants.go` file.
Having a column title saying `Checks` and then seeing some random green character under it might not solve the issue :D

Let me know what you think!

alexkalderimis:
Thanks for your reply @dlvhdr! I can certainly look into adding a more general `useNerdFonts` configuration setting - it is definitely a good idea.

However, I think that customizable column names are both valuable in their own right (for internationalization if nothing else) and the foundation for building more general solutions on.

Personally, I'm already using custom names, and I prefer it

alexkalderimis:
Oh, and actually, I am already using `Checks` for the CI column and with a non-nerd font it does indeed just display coloured boxes - they are actually fine, since the colours contain all the information, so I didn't even notice I was missing anything at first.

For accessibility however, we will need to support custom icon sets of course, so that colourblindness is not an issue.

dlvhdr:
We could replace the icons with text like PASS/FAIL

alexkalderimis:
@dlvhdr - I have created a separate PR (https://github.com/dlvhdr/gh-dash/pull/380) to introduce icon sets. It was clear that doing so is orthogonal to the ability to have custom column headers (something I'd still like of course!).

alexkalderimis:
> We could replace the icons with text like PASS/FAIL

I believe this is best addressed as part of #380 - I would like to keep this PR from growing if possible and to make it limited to the functionality of customizing column headers

dlvhdr:
Hmm I can see some downsides to this approach we can flesh out.
- GitHub themselves don't support localization
- Any config value we add is a contract with our users so any change is potentially a breaking change
- This would require a lot of configuration from the user and I'm not sure it's better than providing a set of defaults with a `useNerdFont` option.

WDYT?

alexkalderimis:
Thanks for your reply @dlvhdr 

> Hmm I can see some downsides to this approach we can flesh out.
> 
>     * GitHub themselves don't support localization

I don't entirely understand the point - this is not a GitHub product, and if I understand it the expected user is someone who wants more control over their interface. I don't think you need to provide translations, but allowing the user to provide their preferred text seems only polite.

> 
>     * Any config value we add is a contract with our users so any change is potentially a breaking change

This is true of all features, of course

> 
>     * This would require a lot of configuration from the user and I'm not sure it's better than providing a set of defaults with a `useNerdFont` option.

I can see the argument - #380 aims to address that approach.

I personally don't have a problem with the level of configuration, but I am just one person.

> 
> 
> WDYT?

I clearly think that this is a good idea - I implemented it for myself and considered that if I wanted it, then others might appreciate it.

I don't mind either way of course how you choose to proceed - it is easy to maintain my own fork/patches for this functionality, if you prefer to close this PR

dlvhdr:
I think I would prefer to implement it with a single property...

alexkalderimis:
That's ok with me - feel free to close this PR and I look forward to your
input on #380

Il sab 1 giu 2024, 12:27 Dolev Hadar ***@***.***> ha scritto:

PR #380, opened 2024-05-30
Add configurable icon sets

alexkalderimis:
# Summary

This introduces the concept of 'icon sets', and the ability to specify
individual icons.

In addition to the default icon set (nerd-fonts), a new base set is
added: `emoji`, which uses widely supported emoji.

This was tested manually using the following configuration:

```yaml
theme:
  icon-set: emoji
```

When run like this, instead of nerd-font icons, standard emoji will be used
instead.

## Images/Videos

Screenshot 2024-05-30 at 7 46 25 PM


config/parser.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/charmbracelet/lipgloss"
 	"github.com/go-playground/validator/v10"
+	emoji "github.com/kyokomi/emoji/v2"
```
alexkalderimis, line 13: This was added to allow users to refer to emoji by name

PR #393, opened 2024-06-17
Issue 392

### LINKED ISSUE #392: [BUG] Publish version/release on pkg.go.dev
I'm trying to package `gh-dash` for OpenBSD to import it in the official ports tree.

The port infrastructure on OpenBSD for Go projects needs to get version/infos from pkg.go.dev => get `MOD_NAME` and latest published version `MOD_VERSION`.

On pkg.go.dev, these infos are not up to date => https://pkg.go.dev/github.com/dlvhdr/gh-dash
```bash
$ go list -m github.com/dlvhdr/gh-dash@latest
github.com/dlvhdr/gh-dash v1.2.3
```

Please could you published the latest version v4.1.2 on pkg.go.dev and for each future release ?

Omnikron13:
# Summary
Added a very large number of '/v4' tokens to import strings, which I belive is the only important thing to actually have proxies pull the correct latest version and therefor automatically resolve #392 on the proxy's side?


dlvhdr (approved):
thank you!

PR #399, opened 2024-06-27
feat(approve): Add PR approval cmd

merkata:
# Summary

Adds the option to approve a PR, this is handy when working in a team and with having some branch protections in place (like require X approvals to merge).

`go run ./gh-dash.go --debug`

## Images/Videos

gh_dash_approve

### DIFF
--- a/ui/components/prsidebar/approve.go
+++ b/ui/components/prsidebar/approve.go
@@ -17,7 +17,7 @@ func (m *Model) approve(comment string) tea.Cmd {
 	task := context.Task{
 		Id:           taskId,
 		StartText:    fmt.Sprintf("Approving pr #%d", prNumber),
-		FinishedText: fmt.Sprintf("pr #%d has been approved with comment: %s", prNumber, comment),
+		FinishedText: fmt.Sprintf("pr #%d has been approved", prNumber),
 		State:        context.TaskStart,
 		Error:        nil,
 	}
--- a/ui/components/prsidebar/approve.go
+++ b/ui/components/prsidebar/approve.go
@@ -1,0 +1,54 @@
+package prsidebar
+
+import (
+	"fmt"
+	"os/exec"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/dlvhdr/gh-dash/v4/ui/components/prssection"
+	"github.com/dlvhdr/gh-dash/v4/ui/constants"
+	"github.com/dlvhdr/gh-dash/v4/ui/context"
+)
+
+func (m *Model) approve(comment string) tea.Cmd {
+	pr := m.pr.Data
+	prNumber := pr.GetNumber()
+	taskId := fmt.Sprintf("pr_approve_%d", prNumber)
+	task := context.Task{
+		Id:           taskId,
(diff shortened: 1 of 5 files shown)


ui/components/prsidebar/approve.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func (m *Model) approve(comment string) tea.Cmd {
+	pr := m.pr.Data
+	prNumber := pr.GetNumber()
+	taskId := fmt.Sprintf("pr_approve_%d", prNumber)
+	task := context.Task{
+		Id:           taskId,
+		StartText:    fmt.Sprintf("Approving pr #%d", prNumber),
+		FinishedText: fmt.Sprintf("pr #%d has been approved with comment: %s", prNumber, comment),
```
dlvhdr, line 20: Let's not include the content of the comment as it can be quite long and have new lines. This will break the UI once rendered

ui/keys/prKeys.go
```diff
@@ -24,6 +25,10 @@ type PRKeyMap struct {
 }
 
 var PRKeys = PRKeyMap{
+	Approve: key.NewBinding(
+		key.WithKeys("!"),
```
dlvhdr, line 29: I think <kbd>v</kbd> would work better here? as in a checkmark?
maybe we can change to <kbd>a</kbd> when we introduce a breaking change.

dlvhdr (commented):
This is great! Thanks for contributing :)
Had some minor comments

ui/keys/prKeys.go
merkata: That's great, changed!

ui/components/prsidebar/approve.go
merkata: Makes sense, adjusted. Thanks!

dlvhdr (approved):

ui/components/prsidebar/prsidebar.go
```diff
@@ -71,6 +73,25 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+		} else if m.isApproving {
+			switch msg.Type {
+
+			case tea.KeyCtrlD:
+				if len(strings.Trim(m.inputBox.Value(), " ")) != 0 {
+					cmd = m.approve(m.inputBox.Value())
+				}
+				m.inputBox.Blur()
+				m.isApproving = false
+				return m, cmd
```
rmacklin, line 87: This seems to require a comment to be entered, or else the approval won't occur. Yet the code here:
https://github.com/merkata/gh-dash/blob/d9033f7855730e536f5a3ecfbd8cb8e9f4c9268b/ui/components/prsidebar/approve.go#L33-L35
seems to indicate that approving with no comment is supposed to be supported. Do we want to support approving without a comment (like this PR itself was approved: https://github.com/dlvhdr/gh-dash/pull/399#pullrequestreview-2149546111)?

rmacklin: I've opened a PR to enable approving without a comment:
- https://github.com/dlvhdr/gh-dash/pull/420

PR #414, opened 2024-08-09
chore: add `Makefile`

kyu08:
# Summary
I added `Makefile`.
This is useful for the person who is new to this repository like me because the commands commonly used tell us how to develop.

If some commands fall off from `Makefile` please let me know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+.PHONY: run
+run:
+		@go run .
+
+.PHONY: run-debug
+run-debug:
+		go run . --debug
+
+.PHONY: test
+test:
+		go test -v ./...


Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: run
+run:
+		@go run .
```
kyu08, line 3: `@` makes the output invisible.

dlvhdr:
thanks @kyu08 , maybe also add a debug command that:
go run . --debug

Also, I have been exploring tools like https://github.com/silenceper/gowatch to quickly develop new features. I want to look at alternatives and it has some bugs

kyu08:
> thanks @kyu08 , maybe also add a debug command that:
> go run . --debug

I added it and combined commits into one.

> Also, I have been exploring tools like https://github.com/silenceper/gowatch to quickly develop new features. I want to look at alternatives and it has some bugs

I used to use [air](https://github.com/air-verse/air) for web dev. I tried to use it with gh-dash but it show nothing. Maybe handling output by cli tool using air may need some hacks.

dlvhdr (approved):

PR #420, opened 2024-08-15
Enable approving a PR without entering a comment

rmacklin:
# Summary

This is a follow-up to https://github.com/dlvhdr/gh-dash/pull/399 to address https://github.com/dlvhdr/gh-dash/pull/399#pullrequestreview-2239389841

That PR added support for approving PRs. However, the code was requiring that a comment be entered, or else the approval wouldn't occur. This seemed unintended because the implementation of the `approve` method had some code to explicitly handle the case where the comment was empty:
https://github.com/dlvhdr/gh-dash/blob/d9033f7855730e536f5a3ecfbd8cb8e9f4c9268b/ui/components/prsidebar/approve.go#L33-L35

It was likely just a copy-paste mistake - the code for leaving a regular comment was aborting if the comment was empty (or only whitespace), which makes sense (but doesn't apply for approvals).

I built this branch locally and ran it and used it to approve a PR without a comment:

image

### DIFF
--- a/ui/components/prsidebar/prsidebar.go
+++ b/ui/components/prsidebar/prsidebar.go
@@ -79,9 +79,11 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 			switch msg.Type {
 
 			case tea.KeyCtrlD:
+				comment := ""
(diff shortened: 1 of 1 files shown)


dlvhdr (approved):
thanks!

Issue #475, opened 2024-11-07
[BUG] Immediate `runtime error: invalid memory address or nil pointer dereference` on launching `gh dash`


cosgroveb:
**Describe the bug**
Immediate `runtime error: invalid memory address or nil pointer dereference` on launching `gh dash`

I believe this is distinct from issue #457 and believe that other respondents on that thread including @carmen-gh and @gotcha are experincing this and not that.

**To Reproduce**
1. Install gh-dash >= 4.6.0
2. `gh dash`
3. Experience immediate nil pointer deref error

**Expected behavior**
No crash on launching `gh dash`

**Screenshots**
N/A... Instead here is the stack trace/exception:

```
gh dash
Caught panic:

runtime error: invalid memory address or nil pointer dereference

Restoring terminal...

goroutine 1 [running]:
runtime/debug.Stack()
        /opt/hostedtoolcache/go/1.23.1/x64/src/runtime/debug/stack.go:26 +0x5e
runtime/debug.PrintStack()
        /opt/hostedtoolcache/go/1.23.1/x64/src/runtime/debug/stack.go:18 +0x13
github.com/charmbracelet/bubbletea.(*Program).Run.func1()
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.1.0/tea.go:519 +0x8b
panic({0xc0a040?, 0x1597c00?})
        /opt/hostedtoolcache/go/1.23.1/x64/src/runtime/panic.go:785 +0x132
github.com/dlvhdr/gh-dash/v4/ui.Model.Update({0x15cc960, {0x0, {0x1094d50, 0x0}, {0x0, 0x61, {{...}, {...}, {...}, {...}, ...}, ...}, ...}, ...}, ...)
        /home/runner/go/pkg/mod/github.com/dlvhdr/gh-dash/v4@v4.7.0/ui/ui.go:549 +0x4909
github.com/charmbracelet/bubbletea.(*Program).eventLoop(0xc000165200, {0x109c5b8?, 0xc0005d6000?}, 0xc000118850)
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.1.0/tea.go:452 +0x810
github.com/charmbracelet/bubbletea.(*Program).Run(0xc000165200)
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.1.0/tea.go:593 +0xa32
[...]

PR #476, opened 2024-11-07
Fix nil pointer deref introduced in 4.6.0 by introducing guard around potentially `nil` currSection

cosgroveb:
This addresses #475

# Summary

I get a runtime error: invalid memory address or nil pointer dereference on launching gh dash when the version is >= 4.6.0.

1. `gh extension install dlvhdr/gh-dash`
2. `gh dash` 
3. See stack trace in #475
4. Remove extension
5. `gh extension install dlvhdr/gh-dash --pin v4.5.4`
6. `gh dash`
7. No crash
8. Remove extension
9. `gh extension install dlvhdr/gh-dash --pin v4.6.0`
10. Crash with stack trace equivalent to the one in #475 
11. `git clone` gh-dash
12. Apply this change
13. `go build` && `go install .`
14. `gh dash`
15. No crash ✅ 

## Images/Videos

ezgif-5-74bd6d56f9

### DIFF
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -546,7 +546,9 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 
 	case execProcessFinishedMsg, tea.FocusMsg:
-		cmds = append(cmds, currSection.FetchNextPageSectionRows()...)
+		if currSection != nil {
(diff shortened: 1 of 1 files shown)


dlvhdr:
ty!

dlvhdr (approved):

gotcha:
@cosgroveb Thanks for working on this. And fixing it for me at least !

PR #483, opened 2024-11-12
docs: consistent formatting in builtin description

alcpereira:
# Summary

While working on the docs of some undocumented features/shortcuts, I noticed that the `?` command helper has some inconsistencies with two of the commands.

I updated the 2 commands descriptions that used capital letters to be all lower case like the other descriptions.

Tested it locally

## Images/Videos

image


dlvhdr (approved):
Thx!

Issue #504, opened 2025-01-10
Support multiple keys for bindings


gabriellanata:
**Is your feature request related to a problem? Please describe.**
I would like to assign multiple keys to an action, but it is currently only possible to assign a single key. Any additional keybindings replace the previous one.

For example, i want both tab and right arrow to go to the next section. However, this doesnt work and only the second key will work:

```
keybindings:
  universal:
    - key: right
      builtin: nextSection
    - key: tab
      builtin: nextSection
```

**Describe the solution you'd like**
I'd like support for multiple keys. The key binding descriptions already include multiple keys by showing them separated by "/". So i propose using the same syntax for adding multiple keys.

```
keybindings:
  universal:
    - key: right/tab
      builtin: nextSection
```

**Describe alternatives you've considered**
Al alternative is to make the original snippet work and having multiple keys be defined separately. This is a perfectly fine alternative.

gabriellanata:
Proposed solution is here: https://github.com/dlvhdr/gh-dash/pull/505

Omnikron13:
In YAML lists should be represented like this:

```
keybindings:
   universal:
    - key: [right, tab]
      builtin: nextSection
```

And then when the YAML is parsed you need to process either the single value or list accordingly.

gabriellanata:
Thanks for the feedback, I updated the PR

PR #505, opened 2025-01-10
Support multiple keys for key bindings

gabriellanata:
# Summary

Adds support for multiple keys for key bindings by allowing specifying a list or a string

Fixes: https://github.com/dlvhdr/gh-dash/issues/504

I have not tested. I am not a go programmer so this is just an implementation suggestion to maybe make it easier for whoever is more familiar with this than me.

### DIFF
--- a/config/parser.go
+++ b/config/parser.go
@@ -111,7 +111,7 @@ type RepoConfig struct {
 }
 
 type Keybinding struct {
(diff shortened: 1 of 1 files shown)


dlvhdr:
I'm getting build errors when trying to run this, are you sure it works?
image

PR #508, opened 2025-01-11
better fix for leak and cleanup of unnecessary string ptr

ldemailly:
(may or not be fully working)

you need to undo the 50ms if you were to merge this

### DIFF
--- a/ui/components/sidebar/sidebar.go
+++ b/ui/components/sidebar/sidebar.go
@@ -83,7 +83,7 @@ func (m *Model) GetSidebarContentWidth() int {
 	if m.ctx.Config == nil {
 		return 0
 	}
-	return m.ctx.Config.Defaults.Preview.Width - 2*m.ctx.Styles.Sidebar.ContentPadding - m.ctx.Styles.Sidebar.BorderWidth
+	return m.ctx.Config.Defaults.Preview.Width - m.ctx.Styles.Sidebar.ContentPadding - m.ctx.Styles.Sidebar.BorderWidth
 }
 
 func (m *Model) ScrollToTop() {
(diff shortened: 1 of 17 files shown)


dlvhdr:
Thanks for this!

ui/components/sidebar/sidebar.go
```diff
@@ -83,7 +83,7 @@ func (m *Model) GetSidebarContentWidth() int {
 	if m.ctx.Config == nil {
 		return 0
 	}
-	return m.ctx.Config.Defaults.Preview.Width - 2*m.ctx.Styles.Sidebar.ContentPadding - m.ctx.Styles.Sidebar.BorderWidth
+	return m.ctx.Config.Defaults.Preview.Width - m.ctx.Styles.Sidebar.ContentPadding - m.ctx.Styles.Sidebar.BorderWidth
```
ldemailly, line 86:
```suggestion
	return m.ctx.Config.Defaults.Preview.Width - 2*m.ctx.Styles.Sidebar.ContentPadding - m.ctx.Styles.Sidebar.BorderWidth
```
fixing search and replace mishap

ui/ui.go
```diff
@@ -960,7 +960,8 @@ func (m *Model) doRefreshAtInterval() tea.Cmd {
 	}
 
 	return tea.Tick(
-		time.Minute*time.Duration(m.ctx.Config.Defaults.RefetchIntervalMinutes),
+		// time.Minute*time.Duration(m.ctx.Config.Defaults.RefetchIntervalMinutes),
+		time.Millisecond*50,
```
ldemailly, line 964: to undo to merge ^

ldemailly:
yw, there are a few more *string you can probably remove but only changed the one I encountered while doing the other change (using * for the shared ProgramContext) - a lot of search and replace basically

ldemailly:
hi, do you plan on merging this?

dlvhdr:
Yes! Need to get to it

PR #510, opened 2025-01-18
build: nix

dlvhdr:
# Summary


bryceberger:
I have some comments / suggestions, in 8f6694d17c53b36ed079c0c80f82f32349f048f3. (Feel free to just add commit to this branch)

- The devshell and build configuration is small enough that it should probably just be inlined into the flake
  - Using `./.` for the source, when in a git repository, already filters out all the `.gitignore`d files. Don't need the big intersection.
- Added language servers (`gopls` + `golangci-lint-langserver`) to the devshell
- Used [flake-utils](https://github.com/numtide/flake-utils) to generate for both mac and linux (both x86 and aarch64)
- Setting `outputs.overlays.default` allows the following use:
```
{
  inputs.gh-dash.url = "github:bryceberger/gh-dash/8f6694d17c53b36ed079c0c80f82f32349f048f3";
  outputs = {
    nixpkgs,
    gh-dash,
    ...
  }: let
    system = "x86_64-linux";
    pkgs = import nixpkgs {
      inherit system;
      overlays = [gh-dash.overlays.default];
    };
  in {
    devShells.${system}.default = pkgs.mkShell {
      # the following gets gh-dash version 4.10, instead of the 4.8.1 currently in nixpkgs
      packages = [pkgs.gh-dash];
    };
  };
}
```

dlvhdr:
Thanks for the tips @bryceberger, I'm still learning nix so it's pretty helpful!

In addition, I'm actually thinking of using https://goreleaser.com/customization/nix/?h=nix to publish it to nixpkgs.

dlvhdr:
I've also pushed your commit here

bryceberger:
[nixpkgs](https://github.com/NixOS/nixpkgs) is different from [nur](https://github.com/nix-community/NUR) --- the former is the official distribution of nixos, the latter is a collection of user packages. Similar to the Arch repos / AUR, if you're familiar with that.

`gh-dash` is already packaged in nixpkgs: https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/gh/gh-dash/package.nix

I'm not entirely familiar with either the NUR or GoReleaser, so can't help you there.

PR #511, opened 2025-01-22
fix: stop compiling static regex in hot path, compile once globally

tjhop:
Previously, these static regexs were being compiled each time the
function is called. The regexs are static, so there's no need to
re-compile them everytime; instead, compile them once globally per
package, and then re-use them when needed. This
reduces the number of allocations made over the life span of the
program; resulting in less active memory use in general, as well as less
garbage collection. Local profiling with parca and a shimmed in pprof
web server show me approx ~20% less in use memory, going from `~8-10MB`
-> `~6-8MB` after changes.

Signed-off-by: TJ Hoplock <t.hoplock@gmail.com>

### DIFF
--- a/ui/components/issuesidebar/activity.go
+++ b/ui/components/issuesidebar/activity.go
@@ -1,7 +1,6 @@
 package issuesidebar
 
 import (
-	"regexp"
 	"sort"
 	"time"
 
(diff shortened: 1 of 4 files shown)


dlvhdr:
nice! thank you!

dlvhdr (approved):
❤️

Issue #515, opened 2025-01-28
Include author-role indicator in the UI for PRs and issues


sideshowbarker:
**Describe the solution you'd like**

For PRs, I’d like the UI to show a _“this is first-time contributor”_ indicator similar in some way to what the GitHub UI shows.

But if we’re going to show that, it seems like we should just go ahead also always some indicator of the author’s role with respect to the repo to which the PR was submitted. I mean, “contributor”, “member”, “owner” etc.

We can get that author-role data with the API — from the value of the `pull_request.author_association` field. The values `NONE`, `FIRST_TIMER`, or `FIRST_TIME_CONTRIBUTOR` all basically indicate “first-time contributor”. See the details at https://michaelheap.com/github-actions-check-permission/.

**Describe alternatives you've considered**
I haven’t considered any alternatives.

**Additional context**

Use case: I’m one of the core reviewers for MDN Web Docs, and each week we get a significant number of people contributing docs patches to MDN for the first time. In those cases, I like to post a special comment welcoming them and thanking them for their first contribution.

Example: https://github.com/mdn/content/pull/37835#issuecomment-2617657110

And when I have a lot of “Needs My Review” PRs to triage, I personally also like to give some priority to the first-time-contributor PRs over PRs from returning contributors — or minimally, I try to make sure those first-time-contributor PRs don’t sit for too long without me or somebody else taking time to comment, at least just to give them some estimate of when we’ll be able to actually review the PR. And I especially like to make sure that those first-time contributor PRs don’t end going unhandled for so long that a project stale bot ends up auto-closing them

**Implementing**

I’d be willing to contribute a patch myself — once I get familiar enough with the code to be able to write up patch.

But if somebody else gets it implemented in the meantime, that’d be great too.

PR #516, opened 2025-01-28
feat: show issue/PR author-role indicators in UI, after author logins

sideshowbarker:
# Summary

This change causes author-role indicators (colored Nerd Font icons) to be shown the dashboard, after the author’s GitHub login. Each indicator shows the author’s role relative to the repo to which they submitted their PR or issue. Fixes https://github.com/dlvhdr/gh-dash/issues/515.

For example, for a PR that’s the first PR the author has contributed to a particular repo, an indicator that the author is a first-time contributor to that repo will be shown.

Other indicators are used for showing that the author has previously contributed to the repo for the PR or issue, or that the author is member or owner of that repo, or a collaborator for that repo.

See https://docs.github.com/en/graphql/reference/enums#commentauthorassociation for details about the possible role (“author association”) values, or see the writeup at https://michaelheap.com/github-actions-check-permission/.

> [!Note]
> The screenshots below show how the author-role-indicator Nerd Font icons would look in the UI after this change.
>
> I don’t feel strongly at all about gh-dash needing to use the particular set of Nerd Font icons I chose in the current patch in this PR branch. So, @dlvhdr: If you have a different set of icons that you’d prefer — to best match with the rest of the existing gh-dash UI design — (or other commenters here have some good alternative suggestions), then I’m 100% fine with some other alternatives might be decided on instead.

I ran `make test` (`go test -v ./...`), to ensure this change didn’t regress any existing tests.

Otherwise, I just tested this change manually by building this PR branch and then switching my config between the compact UI view and the non-compact UI view, and visually inspecting the displayed data to confirm that the author-role indicators are displayed as expected.

## Images/Videos

image

---

image

---

The above two screenshots are using the following set of icons:

image

---

image

---

The above screenshot is using the following alternative set of icons:

image

---

But as I alluded to in the note at the top of the PR description: I don’t feel strongly about either of those sets of icons. Whatever other set might get decided on instead of either of those two would also be fine with me.

### DIFF
--- a/config/parser.go
+++ b/config/parser.go
@@ -71,6 +71,7 @@ type PrsLayoutConfig struct {
 	CreatedAt    ColumnConfig `yaml:"createdAt,omitempty"`
 	Repo         ColumnConfig `yaml:"repo,omitempty"`
 	Author       ColumnConfig `yaml:"author,omitempty"`
+	AuthorIcon   ColumnConfig `yaml:"authorIcon,omitempty"`
 	Assignees    ColumnConfig `yaml:"assignees,omitempty"`
 	Title        ColumnConfig `yaml:"title,omitempty"`
 	Base         ColumnConfig `yaml:"base,omitempty"`
@@ -81,15 +82,16 @@ type PrsLayoutConfig struct {
 }
 
 type IssuesLayoutConfig struct {
-	UpdatedAt ColumnConfig `yaml:"updatedAt,omitempty"`
-	CreatedAt ColumnConfig `yaml:"createdAt,omitempty"`
-	State     ColumnConfig `yaml:"state,omitempty"`
-	Repo      ColumnConfig `yaml:"repo,omitempty"`
-	Title     ColumnConfig `yaml:"title,omitempty"`
-	Creator   ColumnConfig `yaml:"creator,omitempty"`
-	Assignees ColumnConfig `yaml:"assignees,omitempty"`
-	Comments  ColumnConfig `yaml:"comments,omitempty"`
-	Reactions ColumnConfig `yaml:"reactions,omitempty"`
+	UpdatedAt     ColumnConfig `yaml:"updatedAt,omitempty"`
+	CreatedAt     ColumnConfig `yaml:"createdAt,omitempty"`
+	State         ColumnConfig `yaml:"state,omitempty"`
+	Repo          ColumnConfig `yaml:"repo,omitempty"`
+	Title         ColumnConfig `yaml:"title,omitempty"`
+	Creator       ColumnConfig `yaml:"creator,omitempty"`
+	CreatorIcon   ColumnConfig `yaml:"creatorIcon,omitempty"`
+	Assignees     ColumnConfig `yaml:"assignees,omitempty"`
+	Comments      ColumnConfig `yaml:"comments,omitempty"`
+	Reactions     ColumnConfig `yaml:"reactions,omitempty"`
 }
 
 type LayoutConfig struct {
@@ -143,6 +145,15 @@ type Pager struct {
 
 type HexColor string
 
+type ColorThemeIcon struct {
+	NewContributor HexColor `yaml:"newcontributor"   validate:"omitempty,hexcolor"`
+	Contributor    HexColor `yaml:"contributor"      validate:"omitempty,hexcolor"`
+	Collaborator   HexColor `yaml:"collaborator"     validate:"omitempty,hexcolor"`
+	Member         HexColor `yaml:"member"           validate:"omitempty,hexcolor"`
+	Owner          HexColor `yaml:"owner"            validate:"omitempty,hexcolor"`
+	UnknownRole    HexColor `yaml:"unknownrole"      validate:"omitempty,hexcolor"`
+}
+
 type ColorThemeText struct {
 	Primary   HexColor `yaml:"primary"   validate:"omitempty,hexcolor"`
 	Secondary HexColor `yaml:"secondary" validate:"omitempty,hexcolor"`
@@ -164,6 +175,7 @@ type ColorThemeBackground struct {
 }
 
 type ColorTheme struct {
+	Icon       ColorThemeIcon       `yaml:"icon"       validate:"required"`
 	Text       ColorThemeText       `yaml:"text"       validate:"required"`
 	Background ColorThemeBackground `yaml:"background" validate:"required"`
 	Border     ColorThemeBorder     `yaml:"border"     validate:"required"`
@@ -171,6 +183,19 @@ type ColorTheme struct {
 
 type ColorThemeConfig struct {
 	Inline ColorTheme `yaml:",inline"`
+}
+
+type IconTheme struct {
+	NewContributor string `yaml:"newcontributor,omitempty"`
+	Contributor    string `yaml:"contributor,omitempty"`
+	Collaborator   string `yaml:"collaborator,omitempty"`
+	Member         string `yaml:"member,omitempty"`
+	Owner          string `yaml:"owner,omitempty"`
+	UnknownRole    string `yaml:"unknownrole,omitempty"`
+}
+
+type IconThemeConfig struct {
+	Inline IconTheme `yaml:",inline"`
 }
 
 type TableUIThemeConfig struct {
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -172,6 +172,7 @@ theme: # optional, see more info below
 pager:
   diff: less # or delta for example
 confirmQuit: false # show prompt on quit or not
+showAuthorRole: false # show author role ("new", "member", etc.)
```
dlvhdr, line 175: Would this make more sense under `defaults.layout.prs`/`defaults.layout.issues` as well as an option to define it only for a specific section?
e.g.:
```yaml
prSections:
  - title: My OSS Project
    filters: is:open repo:dlvhdr/gh-dash
    layout:
      role:
        hidden: false
```

Not sure, would like to hear your opinion.

ui/constants/constants.go
```diff
@@ -36,4 +36,25 @@ const (
...
+	UnknownRoleIcon    = "󱐡" // U+F1421 nf-md-incognito_circle
+
+	// NewContributorIcon = "󰲡" // U+F0CA1 nf-md-numeric_1_circle_outline
+	// ContributorIcon    = "󰦤" // U+F09A4 nf-md-star_circle_outline
+	// CollaboratorIcon   = "󱟝" // U+F17DD nf-md-crown_circle_outline
+	// MemberIcon         = "󱟝" // U+F17DD nf-md-crown_circle_outline
+	// OwnerIcon          = "󱟜" // U+F17DC nf-md-crown_circle
+	// UnknownRoleIcon    = "󱐡" // U+F1421 nf-md-incognito_circle
+
+	NewContributorColor = "77"  // PaleGreen3
```
dlvhdr, line 54: I suggest putting these under the `ui/theme/theme.go` file.
Maybe you can even reuse existing colors there, as those are configurable in the config file.

config/parser.go
```diff
@@ -313,6 +314,7 @@ func (parser ConfigParser) getDefaultConfig() Config {
 			},
 		},
 		ConfirmQuit: false,
+		ShowAuthorRole: false,
```
dlvhdr, line 317: I'm not seeing where this is used? perhaps you intended to use it in `GetAuthor()`?

dlvhdr (commented):
This is really great! Thanks for the contrib! 🚀

dlvhdr:
forgot to mention, can you also add this to the docs website? Should be easy to find where to add it

README.md
sideshowbarker: The `showAuthorRole/ShowAuthorRole` config bits were left over from when I first started work on the patch and had thought it made sense for whether or not the author roles were shown to be configurable. But as I worked further on implementing the feature, I came to the conclusion that it seemed to make more sense to just always show the author roles — because doing it requires literally just one icon. So it’s unobtrusive, and I couldn’t imagine that most users would find it so useless or annoying as to want to not have it displayed.

But all that said, if you think it instead really should be made configurable — and I’ll update the patch to make it so.

ui/constants/constants.go
sideshowbarker: OK — I’ve moved all the author-role icon colors to the `ui/theme/theme.go` file.

When I first started trying out the icon colors for this, I did actually use some existing colors. But then when I looked at in the built UI, it felt odd because it kind of seemed to unintentionally give the author-role icons the “meanings” from other existing icons and text in the UI; for example, purple meaning “merged”.

So I actually tried to come up with some new colors for the author-role icons that didn’t conflict in that way with the meanings from existing colored icons and text.

All that said, I’m really not a designer and I really don’t know much about trying to come up with color palettes that are complementary with each other etc. So I don’t feel strongly about what the colors should be, and if the ones I came up with in my patch are not the best choices — and if you want to re-use existing colors instead — then it’s 100% fine by me to change them in any way.

sideshowbarker:
> forgot to mention, can you also add this to the docs website? Should be easy to find where to add it

I’ve pushed the other changes I made in response to review comments — but I’ve not yet made the updates to the docs site. I’ll make time to do that after ~12 hours or so from now.

hallerjulian (approved):

README.md
dlvhdr: Yeah U believe it should be configurable, as most repos O work against are work related, and I don't like the extra visual clutter. I do think it's a really addition for OSS maintainers. Gh-dash also supports defining sections for entirely different repos, some of which might not need these icons.

sideshowbarker: Ok, thanks, I understand — I'll update the patch to make it configurable. The question then becomes: Should I make it on by default, or off be default?

dlvhdr: Cool thank you! Hmm regarding default it's a coin toss from me. I guess turn it **on** by default.

sideshowbarker:
OK — I just now pushed an update (amended commit) with the following changes:

- moved both the definitions of the default color values for the icons, and the definitions of the icon strings/characters themselves into the `ui/theme/theme.go` and `config/parser.go` code
- added new config options for both the icon colors and the icons themselves
- added a new `showAuthorIcons` config option for flipping the author-icon display off
- updated the docs with details about all the new config options

https://github.com/dlvhdr/gh-dash/compare/175aa1a506ae7fa69b3b6a1dac301312409a3353..3efc99b7d0e55117b6687e053831a79ca022befd has the delta diff against what you’d previously reviewed.

config/parser.go
```diff
@@ -186,18 +209,20 @@ type UIThemeConfig struct {
 type ThemeConfig struct {
 	Ui     UIThemeConfig     `yaml:"ui,omitempty"     validate:"omitempty"`
 	Colors *ColorThemeConfig `yaml:"colors,omitempty" validate:"omitempty"`
+	Icons  *IconThemeConfig  `yaml:"icons,omitempty" validate:"omitempty"`
```
dlvhdr, line 212: This is a great addition, thanks!

ui/theme/theme.go
```diff
@@ -8,31 +8,61 @@ import (
...
+	SuccessText             lipgloss.AdaptiveColor // config.Theme.Colors.Text.Success
+	WarningText             lipgloss.AdaptiveColor // config.Theme.Colors.Text.Warning
+	ErrorText               lipgloss.AdaptiveColor // config.Theme.Colors.Text.Error
+	NewContributorIconColor lipgloss.AdaptiveColor // config.Theme.Colors.Icon.NewContributor
+	ContributorIconColor    lipgloss.AdaptiveColor // config.Theme.Colors.Icon.Contributor
+	CollaboratorIconColor   lipgloss.AdaptiveColor // config.Theme.Colors.Icon.Collaborator
+	MemberIconColor         lipgloss.AdaptiveColor // config.Theme.Colors.Icon.Member
+	OwnerIconColor          lipgloss.AdaptiveColor // config.Theme.Colors.Icon.Owner
+	UnknownRoleIconColor    lipgloss.AdaptiveColor // config.Theme.Colors.Icon.UnknownRole
+	NewContributorIcon      string                 // config.Theme.Icons.NewContributor
```
dlvhdr, line 28: Let's move the actual icons to `ui/constants/constants.go`

ui/theme/theme.go
```diff
@@ -8,31 +8,61 @@ import (
...
+	ContributorIconColor:    lipgloss.AdaptiveColor{Light: "075", Dark: "075"},
+	CollaboratorIconColor:   lipgloss.AdaptiveColor{Light: "178", Dark: "178"},
+	MemberIconColor:         lipgloss.AdaptiveColor{Light: "178", Dark: "178"},
+	OwnerIconColor:          lipgloss.AdaptiveColor{Light: "178", Dark: "178"},
+	UnknownRoleIconColor:    lipgloss.AdaptiveColor{Light: "178", Dark: "178"},
+	NewContributorIcon:      "󱃱", // U+F10F1 nf-md-hand_heart
+	ContributorIcon:         "󱕾", // U+F157E nf-md-hand_heart_outline
+	CollaboratorIcon:        "", // U+EDA0  nf-fa-handshake_simple
+	MemberIcon:              "", // U+ED9F  nf-fa-handshake_angle
+	OwnerIcon:               "", // U+ED9F  nf-fa-handshake_angle
```
dlvhdr, line 58: I think the `OwnerIcon` should more distinct than the collaborator and member icons. wdyt about "" (\uebb8), "" (\uedcf), "" (\uedd2), "󱇐" (\udb84\uddd0), "󰢏" (\udb84\udf60).

config/parser.go
```diff
@@ -186,18 +209,20 @@ type UIThemeConfig struct {
...
+	PRSections      []PrsSectionConfig    `yaml:"prSections"`
+	IssuesSections  []IssuesSectionConfig `yaml:"issuesSections"`
+	Repo            RepoConfig            `yaml:"repo"`
+	Defaults        Defaults              `yaml:"defaults"`
+	Keybindings     Keybindings           `yaml:"keybindings"`
+	RepoPaths       map[string]string     `yaml:"repoPaths"`
+	Theme           *ThemeConfig          `yaml:"theme,omitempty" validate:"omitempty"`
+	Pager           Pager                 `yaml:"pager"`
+	ConfirmQuit     bool                  `yaml:"confirmQuit"`
+	ShowAuthorIcons bool                  `yaml:"showAuthorIcons"`
```
dlvhdr, line 225: I think this should also be defined in the section config.
This way, I can enable it globally but disable it for the "My PRs" section where it's irrelevant.

dlvhdr (commented):
Thanks for working on this!

config/parser.go
dlvhdr: Something like:

```yaml
prSections:
  - title: Work
    filters: is:open repo:work/project
    layout:
      authorIcon:
        hidden: true
```

sideshowbarker:
OK, thanks — I’ve just pushed an update (amended commit) that includes changes in response to all pending comments.

> I think the `OwnerIcon` should more distinct than the collaborator and member icons. wdyt about "" (\uebb8), "" (\uedcf), "" (\uedd2), "󱇐" (\udb84\uddd0), "󰢏" (\udb84\udf60).

Thanks — those look great. However, I feel a bit that users might potentially be confused somewhat by what the `\uedd2` “nf-fa-user_tie” icon is intended to convey. So, what about instead maybe the `\uedc6` “nf-fa-user_check” icon? Or if not that, another you can think of that might also be appropriate?

Anyway, what I have for now is this:

image

Is that the correct mappings/order for what you had in mind? If not, please lemme know what I should change.

sideshowbarker:
https://github.com/dlvhdr/gh-dash/compare/3efc99b7d0e55117b6687e053831a79ca022befd..e1b03ca6ab6bf1772bf930b79f72f20a856a1616?w=1 is the delta against what you had previously reviewed

dlvhdr (commented):
This looks great, thanks!

dlvhdr (commented):
I think I would change the icons you chose, but for now it's fine.

dlvhdr:
I think that in a future PR I would like to add a label to the sidebar that explains what the icon means

dlvhdr (approved):

PR #523, opened 2025-02-06
feat: Print full URL to PRs and issues

hallerjulian:
# Summary
gh-dash already offers copying the URL ('Y') and opening the PR/issue in GitHub ('o'). When working with dash via an SSH session, none of those shortcuts work.

Improve usability by printing the full URL, making it easy to copy and/or click the link in such cases.

Using dash in the gh-dash repository and navigating through issues/PRs.

## Images/Videos
Issue: 
image

PR:
image

### DIFF
--- a/ui/components/issuesidebar/issuesidebar.go
+++ b/ui/components/issuesidebar/issuesidebar.go
@@ -131,6 +131,8 @@ func (m Model) View() string {
 
 	s.WriteString(m.renderFullNameAndNumber())
 	s.WriteString("\n")
+	s.WriteString(m.renderUrl())
+	s.WriteString("\n")
 
 	s.WriteString(m.renderTitle())
 	s.WriteString("\n\n")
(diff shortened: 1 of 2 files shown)


hallerjulian:
> Hi @hallerjulian! I think that this will make the experience worse for non-SSH users.. Maybe we can detect we're in an ssh connection, and only then display the url? Also, I'm thinking that displaying the URL in the sidebar is much cleaner as it won't clutter the UI. I'm open to other, less intrusive ideas

Detecting an SSH session would be straight forward, I think, by querying for the `SSH_*` variables being set in the environment.

> Also, I'm thinking that displaying the URL in the sidebar is much cleaner as it won't clutter the UI.
I'm open to other, less intrusive ideas

With sidebar, do you refer to the preview tab on the right? Or another type of sidebar?

dlvhdr:
Yes the preview on the right

hallerjulian:
Will raise a new PR in the future, closing this for now.

PR #558, opened 2025-04-01
feat: set proper inputbox limit (pr/issue)

25huizengek1:
# Summary

This PR sets the character limit for all text boxes in gh-dash to 65536 characters.
While it is possible to set this field (CharLimit) to unlimited (-1), 65536 seems like a sensible default: on GitHub, (almost) all text content supplied through their API is limited by 65536 characters.

Currently, the limit for all textboxes in gh-dash is 400 characters. 
This may seem like a bit of a nitpick, though I've run into the issue myself where I was typing away and ran out of space, only to submit the comment, open the issue/pr in the browser and finish my comment by editing it. This is bad UX IMO.

Closes #517
See: https://github.com/orgs/community/discussions/27190

PR #561, opened 2025-04-11
chore: bump bubbles v0.18.0 => v0.21.0

25huizengek1:
## Summary

This PR bumps Bubbles to the newest version. `gh-dash` has been a few versions behind (v0.18.0 => v0.21.0).
The reason I wanted to PR this is that I'm starting to use `gh-dash` more and more, maybe even considering daily-driving it. I'm constantly running into minor annoyances (hence my previous PR (#558)) and I found another one today: as it turns out, proper (GNU) readline support for input fields hasn't been around [for that long in Bubbletea](https://github.com/charmbracelet/bubbles/commit/8a6f531287b77e89f6807f4ba7af668a241b0a9c). Interactions with text inputs in `gh-dash` will become a lot less annoying (especially since Bubbles v0.20.0 has support for focus/blur window events) after merging this.

I'm sorry if my PRs are on the edge of being nitpicky. These are just small personal annoyances, but on the other hand, why not contribute the fix?


dlvhdr:
Keep em coming :)

PR #565, opened 2025-04-24
feat: add redraw, default to Ctrl-L

philpennock:
# Summary

When opening a browser, GNOME things have a tendency to spew messages to stderr which corrupts the gh-dash display.  Nothing made it possible to get a display which then restored the output, the entry area, etc.

A screen redraw command is useful for "something has happened, it shouldn't have, we know it shouldn't have, but that's life, go ahead and redraw please".  This has traditionally been bound to Ctrl-L, so that's what I picked here.

The bubbletea internals appear to have a `repaintMsg{}` but I couldn't figure out how to just send that.  The lightest-weight option I found was to batch together `ExitAltScreen` and `EnterAltScreen`.  This causes a very brief flicker which ... is also traditional and I'm accepting it as "shows positive feedback that the keystroke was acted upon" rather than a bug.

With this, `gh-dash` can be safely used to `o`pen issues in a browser without then needing to quit and restart, because it offers a recovery mechanism.

Pressed `o` on an issue, returned to the dashboard, typed `Ctrl-L` and got a working display once more.

## Images/Videos

Here's a view of a public PR which I've just pressed `o` on, and then returned to my terminal window; look at the bottom and see the mis-display:

screenshot_gh-dash_opened-issue

The effect is worse with a screen-full of issues, as everything gets scrolled up, and the search/query bar gets scrolled up too, never to be redrawn.

And here it is after I've pressed `Ctrl-L` with this feature built:

screenshot_gh-dash_redrawn

### DIFF
--- a/ui/constants/constants.go
+++ b/ui/constants/constants.go
@@ -12,6 +12,7 @@ type KeyMap struct {
 	TogglePreview key.Binding
 	OpenGithub    key.Binding
 	Refresh       key.Binding
+	Redraw        key.Binding
 	PageDown      key.Binding
 	PageUp        key.Binding
 	NextSection   key.Binding
--- a/ui/keys/keys.go
+++ b/ui/keys/keys.go
@@ -18,6 +18,7 @@ type KeyMap struct {
 	LastLine      key.Binding
 	TogglePreview key.Binding
 	OpenGithub    key.Binding
+	Redraw        key.Binding
 	Refresh       key.Binding
 	RefreshAll    key.Binding
 	PageDown      key.Binding
@@ -91,6 +92,7 @@ func (k KeyMap) AppKeys() []key.Binding {
 	return []key.Binding{
 		k.Refresh,
 		k.RefreshAll,
+		k.Redraw,
 		k.TogglePreview,
 		k.OpenGithub,
 		k.CopyNumber,
@@ -135,6 +137,10 @@ var Keys = &KeyMap{
 	RefreshAll: key.NewBinding(
 		key.WithKeys("R"),
 		key.WithHelp("R", "refresh all"),
+	),
+	Redraw: key.NewBinding(
+		key.WithKeys("ctrl+l"),
+		key.WithHelp("Ctrl+l", "redraw display"),
 	),
 	PageDown: key.NewBinding(
 		key.WithKeys("ctrl+d"),
@@ -247,6 +253,8 @@ func rebindUniversal(universal []config.Keybinding) error {
 			key = &Keys.Refresh
 		case "refreshAll":
 			key = &Keys.RefreshAll
+		case "redraw":
+			key = &Keys.Redraw
 		case "pageDown":
 			key = &Keys.PageDown
 		case "pageUp":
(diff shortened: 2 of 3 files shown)


dlvhdr:
hey from tel aviv

philpennock:
Is there anything I can do to help this land?

I couldn't figure out how to get the docs building locally (I have newer hugo than the GitHub workflow wants, which might be related to the build errors), so I didn't include a doc change, but it should be pretty trivial as a one-liner plus title, for someone who can test the changes.

philpennock:
Sure, as long as the functionality is there and can be bound. :)  Though I do think Ctrl-L is dominant enough to be a reasonable default, that's only an opinion and not one which should block the functionality being available in the release binaries.

Thanks!

philpennock:
Although ... I for one need guidance in how I would configure that.  All I can find in the docs is how to bind a key to run arbitrary external commands, with template substitution available.  How would one configure this, please?  I did give everything names, where I could find them.

dlvhdr:
I'm not near a computer right now, but I believe it would be similar to this PR: https://github.com/dlvhdr/gh-dash/pull/399

dlvhdr:
So this is where you would define your new builtin: https://github.com/dlvhdr/gh-dash/blob/main/ui/keys/keys.go?plain=1#L234
And then this is where you would handle it: https://github.com/dlvhdr/gh-dash/blob/main/ui/ui.go?plain=1#L219

In the config you would write:
```yaml
keybindings:
  universal:
    - key: ctrl+l
      builtin: redraw
```

Something like that.

philpennock:
Right, those code changes are already in this PR, but I'll open a second PR shortly with _only_ those code-changes, so you can choose between the two approaches.

The configuration is what I was missing, as I didn't see any docs on binding to a builtin and the docs I did find only mentioned being able to invoke external commands

philpennock:
https://github.com/dlvhdr/gh-dash/pull/569 is just the critical parts of this PR for the key being available, not loaded by default, and showing the config working.  Thank you for that guidance.  Once I went looking to confirm, of course I immediately found the parser.  Sorry, I was not at my best yesterday.

fivemeyestore:
Great implementation of this feature. I particularly like the approach to error handling.

eyesaider:
Excellent contribution to open source! The community will benefit greatly from this.

fivemhostingserver:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

fivemserverhostings:
Interesting solution to this problem! I've been working on similar challenges recently.

godofcodeeyes:
Great implementation of this feature. I particularly like the approach to error handling.

myeyestore:
The documentation in this repository is excellent! Very clear and comprehensive.

eyesgodcode:
Excellent contribution to open source! The community will benefit greatly from this.

eyestorefivem:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

fivemservereyestore:
Impressive project structure! The architecture really demonstrates solid design principles.

PR #579, opened 2025-06-13
fix: update GoReleaser configurations

emmanuel-ferdman:
# Summary
This small PR updates the GoReleaser configuration to resolve the following warning which you can find in the [CI logs](https://github.com/dlvhdr/gh-dash/actions/runs/15363253048/job/43232945167#step:4:23): 
```
DEPRECATED: archives.format should not be used anymore, check https://goreleaser.com/deprecations#archivesformat for more info
```

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -48,7 +48,7 @@ builds:
         goarch: arm
 
 archives:
-  - format: binary
+  - formats: [binary]
     name_template: "gh-dash_{{ .Tag }}_{{ .Os }}-{{ .Arch }}{{if .Arm}}_{{.Arm}}{{end}}"
 checksum:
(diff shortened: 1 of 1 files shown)


dlvhdr (commented):
Thank you!
I wonder if it matters but should it be 'binary' with single quotes?

image

emmanuel-ferdman:
@dlvhdr All versions are syntactically valid (`goreleaser check` returns no errors), but since the rest of the GoReleaser config uses list-style syntax, I updated the `binary` mode section to match that format for consistency (see this example [here](https://goreleaser.com/blog/goreleaser-v2.6/?h=formats#archive-multiple-formats)). Thanks for the review!

dlvhdr (approved):

Issue #676, opened 2025-10-18
[BUG] TUI crash after playing around with keyboard


gioragutt:
**Describe the bug**

Not exactly sure what the issue was, I played around and after one of the keystrokes the TUI crashed.

**Expected behavior**

**Screenshots**

```
➜  notetaker git:(master) gh dash  
Caught panic:

runtime error: slice bounds out of range [9:6]

Restoring terminal...

goroutine 1 [running]:
runtime/debug.Stack()
        /opt/hostedtoolcache/go/1.24.7/x64/src/runtime/debug/stack.go:26 +0x64
runtime/debug.PrintStack()
        /opt/hostedtoolcache/go/1.24.7/x64/src/runtime/debug/stack.go:18 +0x1c
github.com/charmbracelet/bubbletea.(*Program).recoverFromPanic(0x14000170c80, {0x103b651a0, 0x1400165a5d0})
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.3.5/tea.go:810 +0xa0
github.com/charmbracelet/bubbletea.(*Program).Run.func2()
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.3.5/tea.go:601 +0xdc
panic({0x103b651a0?, 0x1400165a5d0?})
        /opt/hostedtoolcache/go/1.24.7/x64/src/runtime/panic.go:792 +0x124
github.com/charmbracelet/bubbles/viewport.Model.visibleLines({0x5c, 0xfffffffffffffffd, {{{0x0, 0x0, 0x0}, {{...}, {...}}, 0x0}, {{0x0, 0x0, ...}, ...}, ...}, ...})
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbles@v0.21.0/viewport/viewport.go:150 +0x204
github.com/charmbracelet/bubbles/viewport.(*Model).GotoBottom(0x140006366f0)
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbles@v0.21.0/viewport/viewport.go:348 +0xd4
github.com/charmbracelet/bubbles/viewport.(*Model).SetContent(0x140006366f0, {0x14000a46000?, 0x2?})
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbles@v0.21.0/viewport/viewport.go:131 +0xb8
github.com/dlvhdr/gh-dash/v4/internal/tui/components/listviewport.(*Model).SyncViewPort(...)
[...]

PR #688, opened 2025-10-31
fix: crash when not enough available height

dlvhdr:
# Summary


dlvhdr:
Closes https://github.com/dlvhdr/gh-dash/issues/676

Issue #716, opened 2025-12-23
Auto complete labels


mchisolm0:
**Is your feature request related to a problem? Please describe.**
For repos with long label names (e.g. label: `area: languages/rust`), it is frustrating to type the entire label out and possibly make a mistake on spelling.

**Describe the solution you'd like**
When adding a label, fetch the labels and give autocomplete suggestions as the user types.

**Describe alternatives you've considered**
I have considered also having a numbered list of the labels and the user can type the number, but in repos with long labels, there are also likely a lot of labels (e.g. vs code, zed, etc.).

**Additional context**
I am not sure this is possible with BubbleTea. If it is and this is something gh-dash would be open to, I am happy to help get a PR up.

<img width="3228" height="1920" alt="Image" src="https://github.com/user-attachments/assets/e88fc973-5865-47bb-82b9-61b96496337f" />

PR #720, opened 2025-12-27
Label completions

mchisolm0:
# Summary
Add a suggestion/completion system for labels. Built on top of https://github.com/dlvhdr/gh-dash/pull/719 to close https://github.com/dlvhdr/gh-dash/issues/716.

I built the extension locally and started it in the terminal. I opened an issue, began typing a label, `ctrl + n` and `ctrl + p` for next/previous, and clicked enter or tab to accept.

## Images/Videos
<video src="https://github.com/user-attachments/assets/5e6bbc0a-90e3-4b14-b128-55ebe03f6e14">Video unable to load</video>

### DIFF
--- a/internal/tui/components/autocomplete/autocomplete.go
+++ b/internal/tui/components/autocomplete/autocomplete.go
@@ -32,7 +32,7 @@ func (s suggestionList) Len() int {
 }
 
 var (
-	NextKey               = key.NewBinding(key.WithKeys(tea.KeyDown.String(), tea.KeyCtrlN.String()), key.WithHelp("↓/ctrl+n", "next"))
+	NextKey               = key.NewBinding(key.WithKeys(tea.KeyDown.String(), tea.KeyCtrlN.String()), key.WithHelp("↓/Ctrl+n", "next"))
 	PrevKey               = key.NewBinding(key.WithKeys(tea.KeyUp.String(), tea.KeyCtrlP.String()), key.WithHelp("↑/Ctrl+p", "previous"))
 	SelectKey             = key.NewBinding(key.WithKeys(tea.KeyTab.String(), tea.KeyEnter.String(), tea.KeyCtrlY.String()), key.WithHelp("tab/enter/Ctrl+y", "select"))
 	RefreshSuggestionsKey = key.NewBinding(key.WithKeys(tea.KeyCtrlF.String()), key.WithHelp("Ctrl+f", "refresh suggestions"))


internal/data/labelapi.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func GetCachedRepoLabels(repoNameWithOwner string) ([]Label, bool) {
+	labelCacheMu.RLock()
+	defer labelCacheMu.RUnlock()
+	labels, ok := repoLabelCache[repoNameWithOwner]
+	return labels, ok
+}
+
+var (
```
dlvhdr: nit: I think it's best to define these at the top, before you use them

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+	return Model{
+		ctx:        ctx,
+		visible:    false,
+		selected:   0,
+		maxVisible: 5,
+		width:      30,
+	}
+}
+
+func (m *Model) Configure(config Config) {
```
dlvhdr:
```suggestion
func (m *Model) SetConfig(config Config) {
```

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+		excludeMap,
+		m.maxVisible,
+	)
+
+	m.selected = 0
+
+	m.visible = len(m.filtered) > 0 && strings.TrimSpace(currentLabel) != ""
+}
+
+func (m *Model) GetSelected() string {
```
dlvhdr: I try to not do `Get` prefixes but yes for `Set` prefixes
```suggestion
func (m *Model) Selected() string {
```

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+}
+
+func (m *Model) Prev() {
+	if len(m.filtered) > 0 {
+		m.selected--
+		if m.selected < 0 {
+			m.selected = len(m.filtered) - 1
+		}
+	}
```
dlvhdr:
```suggestion
	if len(m.filtered) == 0 {
	  return
	}
  m.selected--
	if m.selected < 0 {
		m.selected = len(m.filtered) - 1
	}
```

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+func (m *Model) View() string {
+	if !m.visible || len(m.filtered) == 0 {
+		return ""
+	}
+
+	numVisible := min(len(m.filtered), m.maxVisible)
+
+	var b strings.Builder
+
+	maxLabelWidth := m.width - 4
```
dlvhdr: can you document what the 4 is with a const? or better yet, deduce it from something else, like the padding of a style with `s.GetHorizontalPadding()` or something?

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+
+	for i := 0; i < numVisible && i < len(m.filtered); i++ {
+		label := m.filtered[i]
+		if len(label) > maxLabelWidth {
+			label = label[:maxLabelWidth-3] + "..."
```
dlvhdr: I have a const called `Ellipsis` which can look better here

dlvhdr: let's also deduce the 3

dlvhdr: There's an `ansi.Truncate` function you can use here which takes wide characters into account

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+
+		// Style based on selection
+		if i == m.selected {
+			// Selected row - use inverted colors
+			selectedStyle := lipgloss.NewStyle().
```
dlvhdr: I try to keep all the styles (if they're constant) in `internal/tui/context/styles.go`

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+				Background(m.ctx.Theme.SelectedBackground).
+				Foreground(m.ctx.Theme.InvertedText)
```
dlvhdr: Let's not use inverted text here, it looks quite bad on my theme.
That's because it's usually a darker color (i.e. inverted from a lighter color) and the selection bg _can_ be dark as well. Let's use the `primary` color as that's pretty much guaranteed to work well with the selection bg.

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+			b.WriteString(selectedStyle.Render("❯ " + label))
```
dlvhdr: maybe you can save the selection character in the constants file with the rest of the special characters.

internal/data/labelapi.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"encoding/json"
+	"os/exec"
+	"strings"
+	"sync"
+)
+
+func GetCachedRepoLabels(repoNameWithOwner string) ([]Label, bool) {
+	labelCacheMu.RLock()
```
dlvhdr, line 18: I've never used locks in go before, maybe a mistake on my part, but can you explain why you need it? What's the concurrent flow we're trying to protect against?

The way bubbletea works is that we handle one `tea.Msg` at a time.
Only `tea.Cmd`s run in a separate thread and can cause these kind of issues, and these usually run network calls etc.

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+	// Return top results
+	result := make([]string, 0, maxResults)
+	for i := 0; i < len(matches) && i < maxResults; i++ {
+		result = append(result, matches[i].label)
+	}
+
+	return result
+}
+
+func fuzzyMatch(text, pattern string) bool {
```
dlvhdr: I think it's more robust to go with a ready package for this.
Maybe we can copy what Charm did? Those guys are pretty good with go :D
https://github.com/charmbracelet/crush/blob/main/internal/tui/exp/list/filterable.go?plain=1#L251

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+func (m *Model) Configure(config Config) {
+	if config.MaxVisible > 0 {
+		m.maxVisible = config.MaxVisible
+	}
+	if config.Width > 0 {
+		m.width = config.Width
+	}
+}
+
+func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
```
dlvhdr: I don't see where this is being used. Maybe let's just keep the one in `inputbox`?

internal/tui/components/inputbox/inputbox.go
```diff
@@ -50,8 +53,51 @@ func NewModel(ctx *context.ProgramContext) Model {
...
+}
+
 func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 	var cmd tea.Cmd
+
+	switch msg := msg.(type) {
+	case tea.KeyMsg:
+		if m.autocomplete != nil && m.autocomplete.IsVisible() {
+			switch msg.Type {
+			case tea.KeyUp, tea.KeyCtrlP:
```
dlvhdr: Let's save it in a tea keybinding like you did in `autocomplete.go`

internal/tui/components/inputbox/inputbox.go
```diff
@@ -50,8 +53,51 @@ func NewModel(ctx *context.ProgramContext) Model {
...
+				m.autocomplete.Prev()
+				return m, nil
+			case tea.KeyDown, tea.KeyCtrlN:
+				m.autocomplete.Next()
+				return m, nil
+			case tea.KeyEnter, tea.KeyTab:
```
dlvhdr: Let's add Ctrl+Y, a vim keybinding

internal/tui/components/inputbox/inputbox.go
```diff
@@ -110,3 +165,31 @@ func (m *Model) UpdateProgramContext(ctx *context.ProgramContext) {
 	m.ctx = ctx
 	m.inputHelp.Styles = ctx.Styles.Help.BubbleStyles
 }
+
+func extractCurrentLabel(input string) string {
```
dlvhdr: this seems to be a duplicate. I think you should delete the one in `autocomplete` as it doesn't need to know about the commas.

internal/tui/components/inputbox/inputbox.go
```diff
@@ -50,8 +53,51 @@ func NewModel(ctx *context.ProgramContext) Model {
...
+				return m, nil
+			case tea.KeyEnter, tea.KeyTab:
+				selected := m.autocomplete.GetSelected()
+				if selected != "" {
+					currentInput := m.textArea.Value()
+					currentLabel := extractCurrentLabel(currentInput)
+					newLabel := selected
+
+					if currentLabel != "" {
+						newInput := strings.ReplaceAll(
```
dlvhdr: replaceAll here might remove things the user already typed. Let's only remove the current label until the next comma if it exists?

internal/tui/components/inputbox/inputbox.go
```diff
@@ -50,8 +53,51 @@ func NewModel(ctx *context.ProgramContext) Model {
...
+							currentInput,
+							currentLabel,
+							newLabel,
+						)
+						m.textArea.SetValue(newInput)
+					}
+
+					newValue := m.textArea.Value()
+					if !strings.HasSuffix(newValue, ", ") {
```
dlvhdr: I don't think `inputbox` should be aware of the commas.
I think the parent of it should. Maybe let's just pass a `OnSuggestionSelected` handler from the parent?

internal/tui/components/issueview/issueview.go
```diff
@@ -180,8 +239,14 @@ func (m Model) View() string {
...
 
-	if m.isCommenting || m.isAssigning || m.isUnassigning || m.isLabeling {
+	if m.isCommenting || m.isAssigning || m.isUnassigning {
+		s.WriteString(m.inputBox.View())
+	} else if m.isLabeling {
 		s.WriteString(m.inputBox.View())
+		if m.ac.IsVisible() {
+			s.WriteString("\n")
+			s.WriteString(m.ac.View())
+		}
```
dlvhdr: I suggest we show the autocomplete above the help view.
The help is always shown at the bottom.

internal/tui/components/issueview/issueview.go
```diff
@@ -255,6 +320,7 @@ func (m *Model) getIndentedContentWidth() int {
 func (m *Model) SetWidth(width int) {
 	m.width = width
 	m.inputBox.SetWidth(width)
+	m.ac.SetWidth(width - 10)
```
dlvhdr: Let's just add a minimal padding of like `2`, I think it will work better.

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,270 @@
...
+	return m.visible
+}
+
+func (m *Model) SetPosition(x, y int) {
+	m.posX = x
+	m.posY = y
+}
+
+func (m *Model) SetWidth(width int) {
+	m.width = width
```
dlvhdr:
```suggestion
	m.width = max(0, width)
```

internal/tui/components/issueview/issueview.go
```diff
@@ -97,22 +114,64 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+				m.ac.Hide()
 				return m, nil
 			}
 
 			m.inputBox, taCmd = m.inputBox.Update(msg)
 			cmds = append(cmds, cmd, taCmd)
+
+			currentInput := m.inputBox.Value()
+			existingLabels := m.inputBox.GetAllLabels()
+			m.ac.Filter(currentInput, existingLabels)
```
dlvhdr: I don't think it's clear that `.Filter` will show the autocomplete.
I would expect `.Show` to do that. A bit confusing as I see various calls to `.Hide` but none to `.Show`.

internal/tui/components/issueview/issueview.go
```diff
@@ -339,20 +405,47 @@ func (m *Model) SetIsLabeling(isLabeling bool) tea.Cmd {
...
+	m.inputBox.SetValue(strings.Join(labels, ", "))
+
+	// Reset autocomplete
+	m.ac.Hide()
+	m.ac.SetSuggestions(nil)
 
+	// Trigger label fetching for autocomplete
 	if isLabeling {
+		repoName := m.issue.Data.GetRepoNameWithOwner()
+		if labels, ok := data.GetCachedRepoLabels(repoName); ok {
```
dlvhdr, line 447: If you're willing and want to make this a bit more bullet proof, you can add a 2nd command that will fetch the repo labels in the bg, in case the cache has been stale for 10 minutes or so. Or maybe even always...?

dlvhdr (commented):
Hey! So this is pretty awesome and definitely super useful. Thanks for working on it.
I think in terms of usability it's very nice. I think the code needs some cleanup, mainly for readability, nothing major.
I've also left some design comments etc.
Otherwise, great work!

internal/tui/components/inputbox/inputbox.go
```diff
@@ -74,6 +120,15 @@ func (m Model) View() string {
 		)
 }
 
+func (m Model) ViewWithAutocomplete() string {
+	baseView := m.View()
+	autocompleteView := ""
+	if m.autocomplete != nil {
+		autocompleteView = m.autocomplete.View()
+	}
+	return baseView + "\n" + autocompleteView
```
dlvhdr: Let's also show the autocomplete keys in this component's help menu alongside Ctrl+D etc.

internal/data/labelapi.go
mchisolm0: Based on your comment, it may be a misunderstanding on my part. I was thinking there might be a problem if `SetIsLabeling` was called multiple times. If the user opened and closed the labeling input multiple times quickly, I thought it would try reading the cache each time, multiple of them might fail, and multiple try to fetch/write the labels.

I also thought a command like this would mean it was in a separate thread, but it sounds like you are saying since the function I was worried about is in a `tea.Msg`, it is fine? I have also tested what I was concerned about after removing the locks, and it seems to work without an issue.
```go
// fetchLabels returns a command to fetch repository labels
func (m *Model) fetchLabels() tea.Cmd {
	return func() tea.Msg {
		repoName := m.issue.Data.GetRepoNameWithOwner()
		labels, err := data.FetchRepoLabels(repoName)
		if err != nil {
			return nil
		}
		return RepoLabelsFetchedMsg{Labels: labels}
	}
}
```

Unless I hear differently from you, I am moving forward without the locks since it seems I was just misunderstanding.

dlvhdr: I think you're right on this. Everything that happens in a `tea.Cmd` happens in a goroutine, async.
Since `FetchRepoLabels` happens in a cmd, it can technically happen multiple times in parallel. So let's keep the lock.

internal/tui/components/inputbox/inputbox.go
mchisolm0: I believe you were saying to add this to the select keybinds with Enter and Tab, correct? If so, it should be done, now.

internal/tui/components/inputbox/inputbox.go
mchisolm0: Thank you for the suggestion! 😁 I believe the way it works now, we may even be able to reuse the autocomplete popup in comments, for example, to suggest/autocomplete users after the `@` symbol. I think we will need a different way to replace since I assumed a single line when replacing the labels.

internal/tui/components/issueview/issueview.go
mchisolm0: I added them at the bottom of the suggestions popup because I felt it was too crowded in the inputbox's help. I still feel like it is a bit crowded, so I am open to feedback (see screenshot below).
<details>
<summary>Screenshot with suggestions popup</summary>
CleanShot 2026-01-03 at 16 28 45@2x
</details>

I also added a show/hide toggle, but I put the help for it in the main help with `Ctrl+d` etc. Is there a better keybinding than `Ctrl+h`?

internal/tui/components/issueview/issueview.go
mchisolm0: I set it to `4` currently because `2` let the popup go off the screen.
<details>
<summary>Padding of 2</summary>
Padding of 2
</details>

<details>
<summary>Padding of 4</summary>
Padding of 4
</details>

internal/tui/components/issueview/issueview.go
mchisolm0: I really like this idea. I also took the opportunity to show loading state. It is currently showing in the popup, but I do not think I have it looking great. 
CleanShot 2026-01-03 at 16 42 35@2x

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,294 @@
...
+type FetchState int
+
+func (s suggestionList) Len() int {
+	return len(s.items)
+}
+
+var (
+	NextKey               = key.NewBinding(key.WithKeys(tea.KeyDown.String(), tea.KeyCtrlN.String()), key.WithHelp("↓/ctrl+n", "next"))
+	PrevKey               = key.NewBinding(key.WithKeys(tea.KeyUp.String(), tea.KeyCtrlP.String()), key.WithHelp("↑/Ctrl+p", "previous"))
+	SelectKey             = key.NewBinding(key.WithKeys(tea.KeyTab.String(), tea.KeyEnter.String(), tea.KeyCtrlY.String()), key.WithHelp("tab/enter/Ctrl+y", "select"))
```
dlvhdr, line 37:
```suggestion
	SelectKey             = key.NewBinding(key.WithKeys(tea.KeyTab.String(), tea.KeyEnter.String(), tea.KeyCtrlY.String()), key.WithHelp("tab/enter/ctrl+y", "select"))
```

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,294 @@
...
+		m.filtered = append(m.filtered, match.Str)
+	}
+
+	m.selected = 0
+	// respect suppression: don't auto-show if suppressed
+	if m.hiddenByUser {
+		m.visible = false
+	} else {
+		m.visible = len(m.filtered) > 0
+	}
```
dlvhdr: similar to the code above, maybe can be extracted into `m.UpdateVisible()` func?

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,294 @@
...
+type Model struct {
+	ctx            *context.ProgramContext
+	suggestionHelp help.Model
+	suggestions    []string
+	filtered       []string
+	selected       int
+	visible        bool
+	maxVisible     int
+	posX           int
+	posY           int
```
dlvhdr: I don't think these are being used

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,294 @@
...
+	for i := 0; i < numVisible && i < len(m.filtered); i++ {
+		label := m.filtered[i]
+		if len(label) > maxLabelWidth {
+			label = ansi.Truncate(label, maxLabelWidth-ellipsisWidth, constants.Ellipsis)
+		}
+
+		// Style based on selection
+		if i == m.selected {
+			// Selected row - use inverted colors
+			b.WriteString(m.ctx.Styles.Autocomplete.SelectedStyle.Render(constants.SelectionIcon + label))
```
dlvhdr:
```suggestion
			b.WriteString(constants.SelectionIcon + " " + m.ctx.Styles.Autocomplete.SelectedStyle.Render(label))
```

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,294 @@
...
+		} else {
+			// Non-selected row
+			b.WriteString(" " + label)
```
dlvhdr:
```suggestion
			b.WriteString("  " + label)
```

internal/tui/components/autocomplete/autocomplete.go
```diff
@@ -0,0 +1,294 @@
...
+	case FetchStateLoading:
+		statusView = m.spinner.View() + " Fetching suggestions" + constants.Ellipsis
+	case FetchStateSuccess:
+		statusView = m.ctx.Styles.Common.SuccessGlyph + " Suggestions loaded"
+	case FetchStateError:
+		errMsg := "Failed to fetch suggestions"
+		if m.fetchError != nil {
+			errMsg = m.fetchError.Error()
+		}
+		statusView = m.ctx.Styles.Common.FailureGlyph + " " + errMsg
```
dlvhdr: can we make the text after the glyph use `m.ctx.Theme.FaintText`?

internal/tui/components/inputbox/inputbox.go
```diff
@@ -110,3 +208,14 @@ func (m *Model) UpdateProgramContext(ctx *context.ProgramContext) {
 	m.ctx = ctx
 	m.inputHelp.Styles = ctx.Styles.Help.BubbleStyles
```
dlvhdr, line 204: I think you also need to update the styles for the autocomplete here? not sure why, but the help text isn't using the faint style

internal/tui/components/inputbox/inputbox.go
```diff
@@ -50,8 +66,66 @@ func NewModel(ctx *context.ProgramContext) Model {
...
+					m.textArea.SetCursor(newCursorPos)
+				}
+				m.autocomplete.Hide()
+				return m, nil
+			}
+		}
+		// Allow fetching labels via keybinding at any time
+		if key.Matches(msg, autocomplete.RefreshSuggestionsKey) {
+			return m, func() tea.Msg { return autocomplete.FetchSuggestionsRequestedMsg{} }
+		}
```
dlvhdr: It's kinda a best practice for bubbletea apps to not pass messages for actions that are not asynchronous. So the alternative is to have the issueview itself listen to the keypress and directly create the `m.fetchLabels` cmd.
It's not super important, your call.

internal/tui/components/issueview/issueview.go
dlvhdr: that's a nice detail! I think I would just show the loading indicator if it's actually loading, otherwise I would show nothing. wdyt?

internal/tui/components/issueview/labels.go
```diff
@@ -0,0 +1,131 @@
+package issueview
+
+import (
+	"strings"
+)
+
+// LabelInfo contains information about a label at a specific cursor position
```
dlvhdr, line 7: love the explanations going on here!

internal/tui/context/styles.go
```diff
@@ -230,6 +234,14 @@ func InitStyles(theme theme.Theme) Styles {
...
 		Background(theme.FaintBorder).
 		Foreground(theme.FaintText)
+	s.Autocomplete.PopupStyle = lipgloss.NewStyle().
+		Border(lipgloss.RoundedBorder()).
+		BorderForeground(theme.SecondaryBorder).
+		Foreground(theme.PrimaryText)
+
+	s.Autocomplete.SelectedStyle = lipgloss.NewStyle().
+		Background(theme.SelectedBackground).
+		Foreground(theme.PrimaryText)
```
dlvhdr: can we also make this bold?

dlvhdr (commented):
Looks and behaves really great! Thx for putting so much effort into it!

dlvhdr (commented):
Looks and behaves really great! Thx for putting so much effort into it!

internal/tui/components/autocomplete/autocomplete.go
mchisolm0: I am happy to have this be a lowercase "c", however, I used "C" to match the current choice in the inputbox help text. I do also notice I need to fix the others in the autocomplete help. 

Is the idea to only capitalize `Ctrl` if it is the first option of the keybind? If that's right, I also need to update `ctrl+p` for previous. Just want to make sure I am matching whatever pattern is in place. 

CleanShot 2026-01-04 at 16 46 52@2x

internal/tui/components/autocomplete/autocomplete.go
mchisolm0: I do not see `m.ctx.Theme.FaintText` used, and I get a compiler error that it's undefined. I think you mean `m.ctx.Common.FaintText`, which is what I am using for now.

If that is wrong please let me know. If it is right, no action is needed.

internal/tui/components/inputbox/inputbox.go
mchisolm0: Great catch! I added it to `issueview.go` since that is where `m.ac` is. I originally did a check for `m.ac != nil` before updating the context, but I think since it is in `issueview.go`, `m.ac` will always be defined. Am I thinking about that right?

internal/tui/components/inputbox/inputbox.go
mchisolm0: I would rather take the opportunity to learn the bubbletea way. I'll save breaking the rules for after I know the rules. Thanks for the context. I believe I have fixed this.

internal/tui/components/issueview/issueview.go
mchisolm0: I also rearranged it a bit. wdyt of this?

CleanShot 2026-01-05 at 00 36 14

internal/tui/components/issueview/issueview.go
dlvhdr: Mmm perhaps ctrl+x? but ctrl+h works fine, I don't have a preference

internal/tui/components/issueview/issueview.go
dlvhdr: looks good!

internal/tui/components/autocomplete/autocomplete.go
dlvhdr: huh you're right, let's have it capitalized. I think the `ctrl+n` isn't yet capitalized though

internal/tui/components/inputbox/inputbox.go
dlvhdr: yeah that works great!

dlvhdr:
Looks great @mchisolm0! I'm good with merging this. We can do another pass once bubbletea v2 is released and convert it to a cleaner popup.

internal/tui/components/issueview/issueview.go
mchisolm0: If you don't have a preference, I think ctrl+h is better.

internal/tui/components/autocomplete/autocomplete.go
mchisolm0: They should all be consistent now.

PR #722, opened 2025-12-29
Add mouse support: resize preview, click tabs & rows

hjanuschka:
Adds mouse interactions I've been missing:

**What's new:**
- Drag the left edge of the preview pane to resize it
- Click on tabs to switch sections
- Click on PR/Issue rows to select them
- Scroll wheel works in preview pane and list
- Preview width is saved between sessions (~/.config/gh-dash/state.yml)

**Details:**
- Resize is clamped to 30-150 chars (max 70% of screen)
- Scroll moves 3 lines/items at a time
- Auto-fetches more items when scrolling near bottom

### DIFF
--- a/internal/config/state.go
+++ b/internal/config/state.go
@@ -1,0 +1,77 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+)
+
+const StateFileName = "state.yml"
+
+// State holds runtime state that should persist between sessions
+type State struct {
+	PreviewWidth int `yaml:"previewWidth,omitempty"`
+}
+
+// GetStatePath returns the path to the state file
+func GetStatePath() (string, error) {
+	configDir := os.Getenv("XDG_CONFIG_HOME")
+	if configDir == "" {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
+		configDir = filepath.Join(homeDir, DEFAULT_XDG_CONFIG_DIRNAME)
+	}
+	return filepath.Join(configDir, DashDir, StateFileName), nil
+}
+
+// LoadState loads the state from the state file
+func LoadState() (State, error) {
+	state := State{}
+
+	statePath, err := GetStatePath()
+	if err != nil {
+		return state, err
+	}
+
+	data, err := os.ReadFile(statePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return state, nil // No state file yet, return empty state
+		}
+		return state, err
+	}
+
+	err = yaml.Unmarshal(data, &state)
+	return state, err
+}
+
+// SaveState saves the state to the state file
+func SaveState(state State) error {
+	statePath, err := GetStatePath()
+	if err != nil {
+		return err
+	}
+
+	// Ensure directory exists
+	dir := filepath.Dir(statePath)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return err
+	}
+
+	data, err := yaml.Marshal(state)
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(statePath, data, 0644)
+}
+
+// SavePreviewWidth saves just the preview width to state
+func SavePreviewWidth(width int) error {
+	state, _ := LoadState() // Ignore error, start fresh if needed
+	state.PreviewWidth = width
+	return SaveState(state)
+}
--- a/internal/config/state.go
+++ b/internal/config/state.go
@@ -8,6 +8,7 @@ import (
 )
 
 const StateFileName = "state.yml"
+const DEFAULT_XDG_STATE_DIRNAME = ".local/state"
 
 // State holds runtime state that should persist between sessions
 type State struct {
@@ -15,16 +16,17 @@ type State struct {
 }
 
 // GetStatePath returns the path to the state file
+// State files are stored in $XDG_STATE_HOME (defaults to ~/.local/state)
 func GetStatePath() (string, error) {
-	configDir := os.Getenv("XDG_CONFIG_HOME")
-	if configDir == "" {
+	stateDir := os.Getenv("XDG_STATE_HOME")
+	if stateDir == "" {
 		homeDir, err := os.UserHomeDir()
 		if err != nil {
 			return "", err
 		}
-		configDir = filepath.Join(homeDir, DEFAULT_XDG_CONFIG_DIRNAME)
+		stateDir = filepath.Join(homeDir, DEFAULT_XDG_STATE_DIRNAME)
 	}
-	return filepath.Join(configDir, DashDir, StateFileName), nil
+	return filepath.Join(stateDir, DashDir, StateFileName), nil
 }
 
 // LoadState loads the state from the state file
--- a/internal/tui/components/carousel/carousel.go
+++ b/internal/tui/components/carousel/carousel.go
@@ -233,6 +233,7 @@ func (m Model) View() string {
 // UpdateSize updates the carousel size based on the previously defined
 // items and width.
 func (m *Model) UpdateSize() {
+	// First pass: render items WITHOUT zone markers to allow truncation
 	leftover := m.width
 	itemsContent := ""
 
@@ -243,13 +244,13 @@ func (m *Model) UpdateSize() {
 	lastRight := right
 	for len(m.items) > 0 && leftover > 0 && (left >= 0 || right < len(m.items)) {
 		if currDirection < 0 && left >= 0 {
-			lItem := m.renderItem(left, leftover)
+			lItem := m.renderItemContent(left, leftover)
 			leftover -= lipgloss.Width(lItem)
 			itemsContent = lipgloss.JoinHorizontal(lipgloss.Top, lItem, itemsContent)
 			lastLeft = left
 			left--
 		} else if currDirection > 0 && right < len(m.items) {
-			rItem := m.renderItem(right, leftover)
+			rItem := m.renderItemContent(right, leftover)
 			leftover -= lipgloss.Width(rItem)
 			itemsContent = lipgloss.JoinHorizontal(lipgloss.Top, itemsContent, rItem)
 			lastRight = right
@@ -281,6 +282,7 @@ func (m *Model) UpdateSize() {
 		l -= lipgloss.Width(roIndicator)
 	}
 
+	// Apply truncation to the content (safe because no zone markers yet)
 	if loIndicator != "" {
 		truncate := lipgloss.Width(itemsContent) - l + 1
 		itemsContent = ansi.TruncateLeft(itemsContent, truncate, "")
@@ -294,6 +296,41 @@ func (m *Model) UpdateSize() {
 			itemsContent = ansi.Truncate(itemsContent, l, "")
 			itemsContent = lipgloss.JoinHorizontal(lipgloss.Center, itemsContent,
 				m.styles.Item.Inline(true).Render(constants.Ellipsis))
+		}
+	}
+
+	// Second pass: rebuild with zone markers on items that fit completely
+	// Only add zone markers when we're NOT truncating content
+	needsTruncation := (loIndicator != "" || lipgloss.Width(itemsContent) > l)
+
+	if !needsTruncation && len(m.items) > 0 {
+		// Rebuild with zone markers since no truncation is needed
+		itemsContent = ""
+		leftover = l
+		left = m.cursor
+		right = min(m.cursor+1, len(m.items))
+		currDirection = -1
+
+		for leftover > 0 && (left >= lastLeft || right <= lastRight) {
+			if currDirection < 0 && left >= lastLeft {
+				lItem := m.renderItem(left, leftover)
+				leftover -= lipgloss.Width(lItem)
+				itemsContent = lipgloss.JoinHorizontal(lipgloss.Top, lItem, itemsContent)
+				left--
+			} else if currDirection > 0 && right <= lastRight {
+				rItem := m.renderItem(right, leftover)
+				leftover -= lipgloss.Width(rItem)
+				itemsContent = lipgloss.JoinHorizontal(lipgloss.Top, itemsContent, rItem)
+				right++
+			}
+
+			if left < lastLeft {
+				currDirection = 1
+			} else if right > lastRight {
+				currDirection = -1
+			} else {
+				currDirection = currDirection * -1
+			}
 		}
 	}
 
@@ -376,7 +413,9 @@ func (m *Model) MoveRight() {
 	m.UpdateSize()
 }
 
-func (m *Model) renderItem(itemID int, maxWidth int) string {
+// renderItemContent renders an item without zone marking.
+// Zone markers should be added separately after all truncation operations are complete.
+func (m *Model) renderItemContent(itemID int, maxWidth int) string {
 	var item string
 	if itemID == m.cursor {
 		item = m.styles.Selected.Render(m.items[itemID])
@@ -393,15 +432,21 @@ func (m *Model) renderItem(itemID int, maxWidth int) string {
 		item = ansi.Truncate(r, maxWidth, m.styles.Item.Inline(true).Render(constants.Ellipsis))
 	}
 
-	// Wrap the item in a zone for click detection
-	zoneID := fmt.Sprintf("%s%s%d", TabZonePrefix, m.zonePrefix, itemID)
-	item = zone.Mark(zoneID, item)
-
 	if m.showSeparators && itemID != len(m.items)-1 {
 		return lipgloss.JoinHorizontal(lipgloss.Center, item, m.styles.Separator.Render(m.separator))
 	}
 
 	return item
+}
+
+// renderItem renders an item with zone marking for click detection.
+// This should only be used when the item won't be further truncated.
+func (m *Model) renderItem(itemID int, maxWidth int) string {
+	item := m.renderItemContent(itemID, maxWidth)
+
+	// Wrap the item in a zone for click detection
+	zoneID := fmt.Sprintf("%s%s%d", TabZonePrefix, m.zonePrefix, itemID)
+	return zone.Mark(zoneID, item)
 }
 
 // HandleClick checks if a mouse click event is on a tab and returns the tab index if so
--- a/internal/tui/components/carousel/carousel.go
+++ b/internal/tui/components/carousel/carousel.go
@@ -1,13 +1,18 @@
 package carousel
 
 import (
+	"fmt"
+
 	"github.com/charmbracelet/bubbles/key"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/charmbracelet/x/ansi"
+	zone "github.com/lrstanley/bubblezone"
 
 	"github.com/dlvhdr/gh-dash/v4/internal/tui/constants"
 )
+
+const TabZonePrefix = "carousel-tab-"
 
 // Model defines a state for the carousel widget.
 type Model struct {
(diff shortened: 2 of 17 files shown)


internal/config/state.go
```diff
@@ -0,0 +1,77 @@
...
+
+const StateFileName = "state.yml"
+
+// State holds runtime state that should persist between sessions
+type State struct {
+	PreviewWidth int `yaml:"previewWidth,omitempty"`
+}
+
+// GetStatePath returns the path to the state file
+func GetStatePath() (string, error) {
```
dlvhdr, line 20: I think state files are usually located in `$XDG_STATE_HOME` which defaults to `$HOME/.local/state`

dlvhdr (commented):
yo, this is awesome!
love these changes.

regarding the width state, how would the user reset the width to the one in their config? remove the state file? maybe there should be a way to reset it through a keybind / flag.

I think that before I go over the code in depth, there are some visual bugs that needs to be fixed.
I had to deal with similar issues due to bubblezone, especially because of `MaxHeight` and `MaxWidth`. See https://github.com/lrstanley/bubblezone?tab=readme-ov-file#maxheight-and-maxwidth.

image

hjanuschka:
done, glitch with the maxwidth/height should be fixed


use "P" to reset preview pane width.
also fixed selection issues on larger lists

dlvhdr:
Great work, looks much better!
A couple of issues:
- There are some artifacts when scrolling that the selected line leaves behind.
- The scrolling, at least for me, seems to scroll 6 items per grading. I _think_ it would be best to only scroll 1 item, so each item would be reachable with the scroll-wheel. wdyt?

https://github.com/user-attachments/assets/e4ba6c22-2d79-48b6-807a-57b820cd227a

hjanuschka:
thanks for the video, should all be fixed with latest commit!! 
thanks for the great project!

dlvhdr:
this is great @hjanuschka!! I'm gonna go over it in the next few days

internal/tui/components/carousel/carousel.go
```diff
@@ -382,6 +439,28 @@ func (m *Model) renderItem(itemID int, maxWidth int) string {
 	return item
 }
 
+// renderItem renders an item with zone marking for click detection.
+// This should only be used when the item won't be further truncated.
+func (m *Model) renderItem(itemID int, maxWidth int) string {
+	item := m.renderItemContent(itemID, maxWidth)
+
+	// Wrap the item in a zone for click detection
+	zoneID := fmt.Sprintf("%s%s%d", TabZonePrefix, m.zonePrefix, itemID)
```
dlvhdr: nit: let's reuse this logic in a function , it is repeated below

internal/tui/components/carousel/carousel_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+		// Render the view to set up zones
+		view := c.View()
+		_ = zone.Scan(view)
+
+		// Verify that zone IDs are being used in the rendered output
+		for i := range items {
+			zoneID := fmt.Sprintf("%s%d", TabZonePrefix, i)
+			// The zone should be marked in the view
+			if !contains(view, zoneID) {
```
dlvhdr: this test doesn't seem to have any assertions

internal/tui/components/carousel/carousel_test.go
```diff
@@ -0,0 +1,150 @@
...
+			t.Errorf("Expected -1 for click outside tabs, got %d", result)
+		}
+	})
+}
+
+func contains(s, substr string) bool {
+	return len(s) >= len(substr) && (s == substr || len(s) > 0 && containsHelper(s, substr))
+}
+
+func containsHelper(s, substr string) bool {
```
dlvhdr: why is this function needed? not sure I understand

internal/tui/components/listviewport/listviewport.go
```diff
@@ -121,19 +120,51 @@ func (m *Model) LastItem() int {
 	return m.currId
 }
 
+func (m *Model) SetCurrItem(index int) {
+	if m.NumCurrentItems == 0 {
+		m.currId = 0
```
dlvhdr: can we reuse this in something like a `resetCurrItem()` func?

internal/tui/components/listviewport/listviewport.go
```diff
@@ -121,19 +120,51 @@ func (m *Model) LastItem() int {
...
+		m.currId = index
+		m.topBoundId = 0
+		m.bottomBoundId = 0
+		m.viewport.GotoTop()
+		return
+	}
+
+	if index < m.topBoundId {
+		diff := m.topBoundId - index
+		m.viewport.ScrollUp(diff * m.ListItemHeight)
```
dlvhdr, line 146: I am not sure how but I managed to get to a state where the top item is only shown half way and I can't scroll to the top 2 items.

internal/tui/components/sidebar/sidebar.go
```diff
@@ -7,17 +7,41 @@ import (
...
+	"github.com/dlvhdr/gh-dash/v4/internal/utils"
 )
 
+const (
+	ResizeZoneID     = "sidebar-resize"
+	MinPreviewWidth  = 30
+	MaxPreviewWidth  = 150
+	ResizeHandleChar = "│"
+	ScrollbarWidth   = 1
+	ScrollThumbChar  = "┃"
```
dlvhdr, line 21: I like the style but think it's a bit too invisible. wdyt if we use more bold colors or characters? the thumb's color could be pink for example, or we could use a thicker character for both thumb and track, so you can properly see them.

internal/tui/components/sidebar/sidebar.go
```diff
@@ -43,34 +67,206 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+
+	// Handle drag while resizing
+	if m.isResizing {
+		switch msg.Action {
+		case tea.MouseActionMotion:
+			// Calculate new width based on mouse position
+			// Mouse X is relative to the terminal, sidebar is on the right
+			// New width = ScreenWidth - MouseX
+			newWidth := m.ctx.ScreenWidth - msg.X
+			if newWidth < MinPreviewWidth {
```
dlvhdr: can we use the `max` builtin func here?

internal/tui/components/sidebar/sidebar.go
```diff
@@ -43,34 +67,206 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+				newWidth = MinPreviewWidth
+			}
+			if newWidth > MaxPreviewWidth {
```
dlvhdr: can we use the `min` builtin func here?

internal/tui/components/sidebar/sidebar.go
```diff
@@ -43,34 +67,206 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+				newWidth = MaxPreviewWidth
+			}
+			// Don't let the sidebar take more than 70% of the screen
+			maxWidth := int(float64(m.ctx.ScreenWidth) * 0.7)
+			if newWidth > maxWidth {
```
dlvhdr: same

internal/tui/components/sidebar/sidebar.go
```diff
@@ -43,34 +67,206 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+		width = m.ctx.Config.Defaults.Preview.Width
+	}
 
+	var content string
 	if m.data == "" {
-		return style.Align(lipgloss.Center).Render(
+		// Content style without the left border and scrollbar
+		contentStyle := lipgloss.NewStyle().
+			Height(height).
+			Width(width - 1) // Subtract 1 for the resize handle
```
dlvhdr: although it's over kill, you can use `lipgloss.Width(resizeHandle)`. It will still result in the same simple `1` but it'll not need the comment

internal/tui/components/sidebar/sidebar.go
```diff
@@ -43,34 +67,206 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+		content = contentStyle.Align(lipgloss.Center).Render(
 			lipgloss.PlaceVertical(height, lipgloss.Center, m.emptyState),
 		)
+	} else {
+		// Render scrollbar
+		scrollbar := m.renderScrollbar(height - m.ctx.Styles.Sidebar.PagerHeight)
```
dlvhdr: if we have a scrollbar, we can actually remove the percentage indication at the bottom. Then we don't need `PagerHeight`.

internal/tui/components/sidebar/sidebar.go
```diff
@@ -43,34 +67,206 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
...
+				Render(fmt.Sprintf("%d%%", int(m.viewport.ScrollPercent()*100))))
+
+		mainContent := lipgloss.JoinVertical(lipgloss.Top, viewportContent, pager)
+
+		// Join content and scrollbar
+		content = lipgloss.JoinHorizontal(lipgloss.Top, mainContent, scrollbar)
+	}
+
+	contentHeight := lipgloss.Height(content)
+	if height > contentHeight {
```
dlvhdr, line 177: can you explain why we need this? not sure I understand

internal/tui/components/table/table.go
```diff
@@ -338,3 +357,18 @@ func (m *Model) UpdateTotalItemsCount(count int) {
...
 }
+
+func (m *Model) SetSectionId(id int) {
+	m.sectionId = id
+}
+
+// HandleClick checks if a row was clicked and returns the row index, or -1 if no row was clicked
+func (m *Model) HandleClick(msg tea.MouseMsg) int {
+	for i := range m.Rows {
+		zoneID := fmt.Sprintf("%s%d-%d", RowZonePrefix, m.sectionId, i)
```
dlvhdr: let's reuse

internal/tui/keys/keys.go
```diff
@@ -243,6 +249,8 @@ func rebindUniversal(universal []config.Keybinding) error {
 			key = &Keys.LastLine
 		case "togglePreview":
 			key = &Keys.TogglePreview
+		case "resetPreviewWidth":
```
dlvhdr, line 252: we should also add this to the docs

internal/tui/ui.go
```diff
@@ -614,25 +627,117 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 
 	case tea.MouseMsg:
-		if msg.Action != tea.MouseActionRelease || msg.Button != tea.MouseButtonLeft {
-			return m, nil
+		// Handle sidebar resize and scroll - pass mouse events to sidebar when it's open
+		if m.sidebar.IsOpen {
+			var sidebarCmd tea.Cmd
+			m.sidebar, sidebarCmd = m.sidebar.Update(msg)
+			if sidebarCmd != nil {
```
dlvhdr: I think it's ok to add `nil` cmds

internal/tui/ui.go
```diff
@@ -614,25 +627,117 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+				cmds = append(cmds, sidebarCmd)
+			}
+			// If resizing is in progress, don't process other mouse events
+			if m.sidebar.IsResizing() {
+				return m, tea.Batch(cmds...)
+			}
+			// If sidebar handled a scroll event, return early
```
dlvhdr: why do we need to return early? what bad thing will happen if we don't?

dlvhdr (commented):
had some small nits and questions, but great job!

internal/tui/components/sidebar/sidebar.go
hjanuschka: added a comment, we normalize heights so the resize handle and content align without truncation.

internal/tui/components/sidebar/sidebar.go
hjanuschka: changed the character

internal/tui/components/listviewport/listviewport.go
hjanuschka: Added a guard so when the current item is 0 we reset bounds and `GotoTop()` to keep the top items reachable.

internal/tui/components/carousel/carousel.go
hjanuschka: Extracted a `tabZoneID()` helper and reused it in both places.

internal/tui/components/carousel/carousel_test.go
hjanuschka: added assertion

internal/tui/components/carousel/carousel_test.go
hjanuschka: Dropped the helper and used `strings.Contains` instead.

internal/tui/ui.go
hjanuschka: Dropped the nil guard and always append the sidebar cmd.

internal/tui/ui.go
hjanuschka: Added a comment: we return early to avoid scrolling the main list when the sidebar handles the wheel event.

hjanuschka:
@dlvhdr thanks for the review, feedback addressed!

internal/tui/components/carousel/carousel.go
```diff
@@ -279,6 +299,41 @@ func (m *Model) UpdateSize() {
 		}
 	}
 
+	// Second pass: rebuild with zone markers on items that fit completely
+	// Only add zone markers when we're NOT truncating content
+	needsTruncation := (loIndicator != "" || lipgloss.Width(itemsContent) > l)
+
+	if !needsTruncation && len(m.items) > 0 {
```
dlvhdr: I am trying to understand, why are we not putting zone markers when there's truncation? doesn't this mean items won't be clickable at all, even if they show fully? I would imagine clicks to always be possible, even on items that partially show. At the current state, items are not clickable at all when the overflow indicator shows.

internal/tui/components/sidebar/sidebar.go
dlvhdr: looks good!

internal/tui/components/carousel/carousel.go
```diff
@@ -168,6 +174,18 @@ func WithKeyMap(km KeyMap) Option {
...
 
+// WithZonePrefix sets a unique prefix for zone IDs to avoid conflicts between views.
+func WithZonePrefix(prefix string) Option {
+	return func(m *Model) {
+		m.zonePrefix = prefix
+	}
+}
+
+// SetZonePrefix sets a unique prefix for zone IDs.
+func (m *Model) SetZonePrefix(prefix string) {
```
dlvhdr, line 185: should we also call this for the prview tabs? (Overview/Checks/Activity/...)

internal/tui/ui.go
```diff
@@ -614,25 +627,115 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+					sidebarStartX = m.ctx.ScreenWidth - m.ctx.PreviewWidth
+				} else {
+					sidebarStartX = m.ctx.ScreenWidth - m.ctx.Config.Defaults.Preview.Width
+				}
+			}
+			if msg.X < sidebarStartX && msg.Y >= common.TabsHeight {
+				section := m.getCurrSection()
+				if section != nil {
+					if msg.Button == tea.MouseButtonWheelUp {
+						section.PrevRow()
```
dlvhdr, line 667: I see we are still scrolling 3 rows at a time. It's not very bad, so not a blocker, but if you want to improve it we can probably do some kind of logic around that.
I wish bubbletea would expose how much scroll happened but we can deduce that from the amount of scroll events we received. So if we received 3 scroll events in quick succession then scroll one line. You would have to keep a state of `lastScrollEvent` or something. Then if we only received one scroll event and some time passed without another (say 50ms) then scroll a line as well.

internal/tui/components/carousel/carousel.go
hjanuschka: Fixed! Changed to a single-pass rendering approach that applies zone markers before truncation.

internal/tui/components/carousel/carousel.go
hjanuschka: Done. Added SetZonePrefix() call to the preview tabs carousel in UpdateProgramContext to match the pattern in the main tabs component. This ensures unique zone IDs and prevents potential conflicts between different carousel instances.

internal/tui/ui.go
hjanuschka: Implemented smart scroll handling. Changed to scroll 1 row per scroll event instead of jumping multiple rows. This gives responsive scrolling behavior where rapid wheel scrolling generates multiple events for fast scrolling, while slow scrolling  produces fewer events.  but i am afraid that is the best we can get now!

internal/tui/ui.go
```diff
@@ -614,25 +628,124 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			}
+			if msg.X < sidebarStartX && msg.Y >= common.TabsHeight {
+				section := m.getCurrSection()
+				if section != nil {
+					// Smart scroll detection: scroll 1 row per event
+					// Track scroll event timing to ensure responsive scrolling.
+					// Each scroll event scrolls 1 row, which gives smooth behavior:
+					// - Rapid wheel scrolling generates multiple events, so scrolling is fast
+					// - Slow scrolling generates fewer events, so scrolling is slower
+					m.lastScrollTime = time.Now()
```
dlvhdr: hmm this seems not to be used. I think you wanted to use it in the if below?

internal/tui/ui.go
```diff
@@ -614,25 +628,124 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+
+					// Always scroll 1 row per event (rapid events scroll more, slow events scroll less)
+					if msg.Button == tea.MouseButtonWheelUp {
+						section.PrevRow()
+					} else {
+						section.NextRow()
+						prevRow := section.CurrRow()
+						nextRow := section.NextRow()
```
dlvhdr, line 670: are these needed here? they look a bit fishy and it seems that scrolling down actually scrolls 6 rows.
Maybe you just wanted to call `NextRow` once and save it in a var?

internal/tui/ui.go
hjanuschka: You're right, it was leftover from an earlier iteration. Removed the unused field and assignment.

internal/tui/ui.go
hjanuschka: Good catch! Fixed. Now we get prevRow before calling NextRow(), so it only scrolls once per event.

schickling:
Really excited to see this work! I recently started using gh-dash as part of my dev workflow (embedded in a zellij tab for PR monitoring), and as someone who is just getting started with gh-dash, mouse support would make a huge difference in lowering the barrier to getting comfortable with the tool. Being able to click around to explore rather than memorizing keybindings first is very welcome.

Just wanted to chime in with some encouragement — are there any remaining blockers or things the community could help with to get this across the finish line?

dlvhdr:
Just need to polish it a bit. I'm intending to merge this soon, just need to find the time.

drod3763:
I'd love to see this merged. Anything to help get this over the finish line?

dlvhdr:
I've just merged bubbletea v2 so this needs to be rebased

orenmizrahi:
is this coming in the next release ? does anyone know the release dates/roadmap ?

drod3763:
This pr has been open for a while now, aside from the merge conflicts, is there anything else that needs to be done here?

m3ac-AllbrittenJ:
would love to see this merged and released too :) I keep forgetting dash doesn't support this while I'm in herdr

gorillamoe:
> would love to see this merged and released too :) I keep forgetting dash doesn't support this while I'm in herdr

How dare you mention herdr here! This is only free-range, grass-fed, non-gmo and jon-AI!

Mane087:
I'm waiting for this new feature 🚀

PR #723, opened 2025-12-29
Add theming system

hjanuschka:
## Summary
- add built-in themes and theme selector
- persist selected theme between sessions
- ensure themed backgrounds render consistently across the UI

Bildschirmfoto 2025-12-29 um 23 05 43

Bildschirmfoto 2025-12-29 um 23 06 04

Bildschirmfoto 2025-12-29 um 23 07 21

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -8,9 +8,11 @@ import (
 	"fmt"
 	slog "log"
 	"os"
+	"os/signal"
 	"runtime"
 	"runtime/debug"
 	"runtime/pprof"
+	"syscall"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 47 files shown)


dlvhdr:
Hey @hjanuschka! This is a great idea.
I think the approach we should take for theming is to use https://github.com/lrstanley/bubbletint. @lrstanley is working on supporting the newest bubbletea version - v2, so this library is well maintained.
His library supports all the themes you can think about. What we would have to do is to only map generic colors to `dash`'s theme and we get the rest for free.
As for the theme selector, it looks great, but kinda bummer that it doesn't show the ui while you're switching between themes.
Bubbletea v2 which supports popups isn't out yet, but we can show the modal where the help section is for now.
It'd be awesome if you can continue working on it!

dlvhdr:
Closing as no activity or responses were made. Looks vibe coded and my approach would be very different as suggested. Feel free to open a new PR with feedback in mind.

Issue #718, opened 2025-12-26
[BUG] Crash when pressing 'w' to focus preview pane on section with external repos


EmZod:
## Description
gh-dash crashes immediately when pressing `w` to focus the preview pane on a section that contains PRs from external/public repositories (repos the user doesn't own).

## Steps to Reproduce
1. Add a section with multiple external repos:
```yaml
prSections:
  - title: Starred Repos
    filters: >-
      is:open
      repo:dlvhdr/gh-dash repo:charmbracelet/glow repo:cline/cline
```
2. Launch gh-dash
3. Navigate to the "Starred Repos" section
4. Select any PR
5. Press `w` to focus the preview pane
6. **Crash occurs** - no output to debug log, appears to be a low-level panic

## Expected Behavior
Preview pane should receive focus and allow scrolling/interaction.

## Environment
- gh-dash version: v4.20.1
- OS: macOS (Darwin, arm64)
- Terminal: Warp

## Notes
- Pressing `w` works fine on sections with owned repos (e.g., "My Pull Requests")
- The crash happens even with small sections (~3 repos)
- Preview pane toggle (`p`) works fine
- No debug output before crash

EmZod:
Thanks for the fix guys! Loving dash

PR #725, opened 2026-01-01
fix: crash when pressing “w” to watch checks

sideshowbarker:
# Summary

This change fixes a case where the `watchChecks` function was creating an empty `prrow.Data{}` struct, and calling `GetStatusChecksRollup()` on it – which led to a nil-pointer dereference when accessing `pr.Data.Primary.Commits.Nodes`.

The change fixes that by not doing the empty-`prrow.Data{}`-struct thing — and instead, switching to using the actual PR data from the current row. It also adds a nil check to `GetStatusChecksRollup()`, for further safety.

Fixes https://github.com/dlvhdr/gh-dash/issues/718

Added unit tests for `GetStatusChecksRollup()`.

### DIFF
--- a/internal/tui/components/prrow/prrow.go
+++ b/internal/tui/components/prrow/prrow.go
@@ -92,6 +92,9 @@ func (pr *PullRequest) renderState() string {
 }
 
 func (pr *PullRequest) GetStatusChecksRollup() checks.CommitState {
+	if pr.Data == nil || pr.Data.Primary == nil {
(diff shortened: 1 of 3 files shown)


dlvhdr (approved):
amazing, thanks for the help!

sideshowbarker:
> amazing, thanks for the help!

Continued thanks for creating & maintaining (and ongoing refining) of such a great app, that’s a genuine delight to hack on~

PR #727, opened 2026-01-02
fix: preserve label colors when updating issue labels

### LINKED ISSUE #721: [BUG] Labels have wrong color after submitting
**Describe the bug**
After submitting labels, the issue gets updated with the new labels but the labels have the wrong colors.

**To Reproduce**
1. Open `gh-dash`
2. Switch to "Issues" view
3. Start labelling with `shift + L`
4. Type a label
5. Submit the label with `ctrl + d`
6. Observe the labels with incorrect colors

**Expected behavior**
Labels should have the correct colors when shown in the issue.

**Screenshots**

Image

 - OS: Mac

majiayu000:
Fixes #721

## Changes
- Create a map to store existing label colors before the gh CLI update
- Use the stored colors when building the returned labels after update

### DIFF
--- a/internal/tui/components/issueview/label.go
+++ b/internal/tui/components/issueview/label.go
@@ -36,6 +36,11 @@ func (m *Model) label(labels []string) tea.Cmd {
 		labelsMap[label] = true
 	}
 
+	existingLabelsColorMap := make(map[string]string)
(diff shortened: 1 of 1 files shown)


dlvhdr (approved):
Thanks that's great! What we really need to do is fetch the labels again though since the new labels won't have color :(

PR #724, opened 2025-12-31
feat: show requested reviewers in PR sidebar

sideshowbarker:
# Summary

This change causes a new “reviewers:” line to be displayed in the PR preview pane, just below the author information.

The “reviewers:” display includes:

- Review-status icon:
  - “waiting” icon for pending reviews
  - “comment” icon, for reviewers who have commented
  - “success” icon for approved reviews
  - “failure” icon for “changes requested”
- Reviewer GH name, with @ prefix for users, no prefix for teams
- Reviewer-type indicators:
  - “team icon for reviews requested from GH teams
  - “owner” icon for code owners

In the case of reviews requested from bots: For GH usernames containing `bot`, we skip the code-owner icon annotation — since the GH name already makes clear that the reviewer is a bot.

The implementation fetches reviewer data via GraphQL using a union type to handle the different reviewer types (User, Team, Bot, Mannequin) — with helper methods to extract display names and determine reviewer types.

Closes #685

## How did you test this change?

Added unit tests in internal/tui/components/prview/reviewers_test.go:

- Empty review requests
- Single user awaiting review
- User who commented
- Code-owner display
- Team reviewer display
- Bot reviewer (no extra annotation)
- Multiple reviewers
- Reviewer who approved
- Reviewer who requested changes
- Mix of pending reviews and completed reviews

### “waiting for review” + “changes requested” + “made comments” reviewers

image

### “waiting for review” + “approved” reviewers

image

### “team” + “bot”  reviewers

image

### “code owner”  reviewers

image

PR #728, opened 2026-01-02
feat: fetch all/more reviewers + show “suggested” (code owner) reviewers too

sideshowbarker:
# Summary

This change implements two follow-ups in response to #724 review comments:

https://github.com/dlvhdr/gh-dash/pull/724#discussion_r2656612589
https://github.com/dlvhdr/gh-dash/pull/724#discussion_r2656616908

Specifically:

- Increases # of reviewers limit from last 5 reviewers to 100 reviewers
- Shows “Loading…” state while enriched data is being fetched
- Shows “suggested” reviewers (= code owners), even if not yet requested
- Shows also those reviewers who only commented (not just approved/rejected)

Updated internal/tui/components/prview/reviewers_test.go with new tests.

### DIFF
--- a/internal/tui/components/issueview/label.go
+++ b/internal/tui/components/issueview/label.go
@@ -36,6 +36,11 @@ func (m *Model) label(labels []string) tea.Cmd {
 		labelsMap[label] = true
 	}
 
+	existingLabelsColorMap := make(map[string]string)
+	for _, label := range m.issue.Data.Labels.Nodes {
+		existingLabelsColorMap[label.Name] = label.Color
+	}
+
 	for _, label := range m.issue.Data.Labels.Nodes {
 		if _, ok := labelsMap[label.Name]; !ok {
 			commandArgs = append(commandArgs, "--remove-label")
@@ -56,7 +61,10 @@ func (m *Model) label(labels []string) tea.Cmd {
 
 		returnedLabels := data.IssueLabels{Nodes: []data.Label{}}
 		for _, label := range labels {
-			returnedLabels.Nodes = append(returnedLabels.Nodes, data.Label{Name: label})
+			returnedLabels.Nodes = append(returnedLabels.Nodes, data.Label{
+				Name:  label,
+				Color: existingLabelsColorMap[label],
+			})
 		}
 		return constants.TaskFinishedMsg{
 			SectionId:   m.sectionId,
--- a/internal/tui/keys/keys.go
+++ b/internal/tui/keys/keys.go
@@ -171,7 +171,7 @@ var Keys = &KeyMap{
 		key.WithHelp("?", "help"),
 	),
 	Quit: key.NewBinding(
-		key.WithKeys("q", "esc", "ctrl+c"),
+		key.WithKeys("q", "ctrl+c"),
 		key.WithHelp("q", "quit"),
 	),
 }


internal/tui/components/prview/prview.go
```diff
@@ -330,10 +330,20 @@ type reviewerItem struct {
...
 
 func (m *Model) renderRequestedReviewers() string {
-	reviewRequests := m.pr.Data.Primary.ReviewRequests.Nodes
-	reviews := m.pr.Data.Primary.Reviews.Nodes
+	if !m.pr.Data.IsEnriched {
+		return lipgloss.JoinVertical(
+			lipgloss.Left,
+			m.ctx.Styles.Common.MainTextStyle.Underline(true).Bold(true).Render("Reviewers"),
+			"",
+			m.ctx.Styles.Common.WaitingGlyph+" Loading...",
```
dlvhdr: I think let's use the faint text style on the `Loading...` text and then join it with the glyph with `lipgloss.JoinHorizontal`. That way it's less eye catching

internal/tui/components/prview/prview.go
```diff
@@ -388,22 +398,44 @@ func (m *Model) renderRequestedReviewers() string {
...
+			continue
+		}
+		shownReviewers[login] = true
+
+		reviewerStr := lipgloss.JoinHorizontal(lipgloss.Top,
+			faintStyle.Render(constants.OwnerIcon), " ",
+			faintStyle.Render("@"+login),
+		)
+
+		reviewerItems = append(reviewerItems, reviewerItem{text: reviewerStr, hasOwnerIcon: true})
```
dlvhdr: There's a small visual bug where there's an extra space before the comma in the reviewers list.
image

sideshowbarker: > There's a small visual bug where there's an extra space before the comma in the reviewers list.

Yup, just now pushed a fix for that one

internal/tui/components/prview/prview.go
sideshowbarker: > I think let's use the faint text style on the `Loading...` text and then join it with the glyph with `lipgloss.JoinHorizontal`. That way it's less eye catching

Made it so.

dlvhdr:
@sideshowbarker I did a bit more testing and it seems there's a small bug.
When reviewers leave several reviews we show the last one's state. This can lead to a wrong conclusion.
For example, if a reviewer approved a PR, then did another review with a "COMMENT" state, we will show the comment icon instead of the check mark icon.
The solution is that we shouldn't override existing review states with a COMMENT state.
Also, do we need to make sure the reviews are ordered by creation date or does that already happen?

PR for example: https://github.com/charmbracelet/bubbles/pull/639

sideshowbarker:
> @sideshowbarker I did a bit more testing and it seems there's a small bug. When reviewers leave several reviews we show the last one's state. This can lead to a wrong conclusion. For example, if a reviewer approved a PR, then did another review with a "COMMENT" state, we will show the comment icon instead of the check mark icon. The solution is that we shouldn't override existing review states with a COMMENT state.

hmm yeah, you’re right — when we are looping there, we just need to continue for any COMMENT we reach when the previous state is a non-comment. So I will make it do that

> Also, do we need to make sure the reviews are ordered by creation date or does that already happen?

Pretty sure that is how they are ordered already. I think the only problem with that is, for a comment which in creation order comes in after a non-comment, we just want that to be a no-op (like discussed above). But for anything else, we do want to just take them as-is as they are — in creation-date order, 

But I will test it and make sure.

sideshowbarker:
> The solution is that we shouldn't override existing review states with a COMMENT state.

Made it so.

> Also, do we need to make sure the reviews are ordered by creation date or does that already happen?

So after looking and thinking, I believe we’re doing the right thing — with ordering y UPDATED_AT. Treating the “follow up” comments as more-significant than the actual actions just had kind of thrown a wrench in that. But with the follow-up-comment handling made right now, I think the basic handling/ordering is how we want it to be.

dlvhdr (approved):
looks great! thx again for the work!

dlvhdr (approved):

PR #732, opened 2026-01-05
fix: multiple crash-inducing bugs and resource leaks

sideshowbarker:
# Summary

- BranchData.GetUpdatedAt(): add nil check for LastUpdatedAt pointer
- BranchData.GetRepoNameWithOwner(): add empty check for Remotes slice
- reposection/prssection/issuessection GetCurrRow(): validate GetCurrItem()
  index is within bounds before accessing slice
- cmd/root.go: fix file handle leak by assigning opened file to loggerFile
  so it gets returned and closed by caller
- watchChecks.go: use captured values instead of calling GetCurrRow() again
  inside goroutine (race condition)
- prapi.go: reuse cachedClient instead of creating new GraphQL client on
  every FetchPullRequest call (resource waste + variable shadowing)
- markdownRenderer.go: handle error from glamour.NewTermRenderer with
  fallback renderer instead of dereferencing nil pointer

Added internal/tui/components/branch/data_test.go

### DIFF
--- a/internal/data/prapi.go
+++ b/internal/data/prapi.go
@@ -381,10 +381,14 @@ type PullRequestsResponse struct {
 	PageInfo   PageInfo
 }
 
-var client *gh.GraphQLClient
+var (
+	client       *gh.GraphQLClient
+	cachedClient *gh.GraphQLClient
+)
 
 func SetClient(c *gh.GraphQLClient) {
 	client = c
+	cachedClient = c
 }
 
 func FetchPullRequests(query string, limit int, pageInfo *PageInfo) (PullRequestsResponse, error) {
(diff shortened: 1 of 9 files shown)


internal/data/prapi.go
```diff
@@ -381,7 +381,10 @@ type PullRequestsResponse struct {
 	PageInfo   PageInfo
 }
 
-var client *gh.GraphQLClient
+var (
+	client       *gh.GraphQLClient
+	cachedClient *gh.GraphQLClient
```
dlvhdr, line 386: what's the difference between these?
The `SetClient` method is used in tests so maybe you just need to set `cachedClient` in it?

sideshowbarker: Yes — thanks for catching that. Given:

  - `client` — used for `FetchPullRequests` (plural — list queries); no caching
  - `cachedClient` — used for `FetchPullRequest` (singular — single PR enrichment); has 5-minute cache
  
…and that I had the tests call `SetClient()`, then that was only setting `client` — so `FetchPullRequest` would still create its own client.

> so maybe you just need to set `cachedClient` in it?

Yeah. Made it that.

Issue #141, opened 2022-06-21
[Feature Request] Support notifications section


mtwig:
# Feature or problem?
This is a feature request to manage github notifications. Github notifications are filtered similarly to issues, and prs.

# Solution I'd like:
* An additional set of panels for notifications, and notification filters.
* Keyboard shortcut to mark a notification as done, subscribe, or bookmark 
  * Reading through notifications should be fast, move to next after taking an action 
* Optional shortcut to mark all notifications in the filter at once. 

# Alternates I've considered
Using an alternative plugin to mark all notifications as read.

dlvhdr:
Yeah I would also like that :)

PR #738, opened 2026-01-08
Notifications dashboard

sideshowbarker:
# Summary

This change adds a new Notifications view to dash that displays GH notifications — allowing you to triage you GH "inbox" within dash. Fixes https://github.com/dlvhdr/gh-dash/issues/141

- View unread notifications with type icons (PR, Issue, Discussion, Release)
- Mark notifications as done (d) or read (m)
- Mark all notifications as done (D) or read (M)
- Unsubscribe from notification threads (u)
- Bookmark notifications to keep them visible after marking read (b)
- Open notifications in browser (o) or view in sidebar (Enter)
- Sort by repository (S)
- Filter by repo:owner/name and is:unread/read/all/done
- New-comment count indicator for PR/Issue notifications
- Auto-scroll to latest comment when viewing
- Active view indicator styled with bold text and gold color for notifications

Bookmark system:
- Local storage in ~/.config/gh-dash/bookmarks.json
- Bookmarked items appear in default view even when read
- Explicit "is:unread" search excludes bookmarked+read items

The implementation follows existing patterns from PR and Issue views, but in this case with requiring you to explicitly/intentionally initiate the "view" action — to prevent accidental "read" marking.

Added new tests:

- tui/components/notificationrow/data_test.go
- notificationssection/filters_test.go

## Images/Videos

image

image

image

### DIFF
--- a/internal/tui/ui_test.go
+++ b/internal/tui/ui_test.go
@@ -25,8 +25,8 @@ import (
 	"github.com/dlvhdr/gh-dash/v4/internal/tui/context"
 	"github.com/dlvhdr/gh-dash/v4/internal/tui/keys"
 	"github.com/dlvhdr/gh-dash/v4/internal/tui/markdown"
-	"github.com/dlvhdr/gh-dash/v4/internal/tui/theme"
 	"github.com/dlvhdr/gh-dash/v4/internal/tui/testutils"
+	"github.com/dlvhdr/gh-dash/v4/internal/tui/theme"
 )
 
 func TestFullOutput(t *testing.T) {


mchisolm0:
Looks awesome! Wanted to mention that it seems the `?` help menu seems to needs a `s to switch to notifications` and move the `s to switch to prs` to the notifications context.

<details><summary>Video</summary>
<video src="https://github.com/user-attachments/assets/b4a6bad7-84d9-4365-9b23-73628be7883e">Unable to load video</video>
</details>

sideshowbarker:
> Looks awesome! Wanted to mention that it seems the `?` help menu seems to needs a `s to switch to notifications` and move the `s to switch to prs` to the notifications context.

Thanks for catching that~ I’ve fixed it now, and pushed an update to this branch

dlvhdr:
Hey man, I did a quick mockup of how I would like this to look.
If you don't have time to take it there, no worries, we can always do it later.
But it's not too different.

Things to note:
- Every row should have an icon. Rows that don't look out of place. Let's try and find one that fits each notification type.
- Rows are 3 lines high to fit all the info clearly (I might make PRs / issues 3 line high as well).
- The activity column I am not sure about, since right now it shows just the number of new comments but maybe we can show there additional info like the content of a comment. Or we can do what github does on its mobile app which is to show the comment content where I just wrote "@someone commented".

image

dlvhdr:
Here's how I would show the preview pane:
image

sideshowbarker:
> Rows are 3 lines high to fit all the info clearly (I might make PRs / issues 3 line high as well).

What if the title is too long to fit in one line? e.g., if the user is showing the Preview?
- If we wrap the title, the we end up with 4 lines rather than 3
- if we want to keep it at exactly 3 lines, we have to truncate/elide the title

dlvhdr:
> > Rows are 3 lines high to fit all the info clearly (I might make PRs / issues 3 line high as well).
> 
> 
> 
> What if the title is too long to fit in one line? e.g., if the user is showing the Preview?
> 
> - If we wrap the title, the we end up with 4 lines rather than 3
> 
> - if we want to keep it at exactly 3 lines, we have to truncate/elide the title
> 
> 

I think it's fine to put ellipsis there

internal/data/bookmarks.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		store.filePath = store.getBookmarksFilePath()
+		store.load()
+		bookmarkStore = store
+	})
+	return bookmarkStore
+}
+
+func (s *BookmarkStore) getBookmarksFilePath() string {
+	configDir := os.Getenv("XDG_CONFIG_HOME")
```
dlvhdr: take a look at how https://github.com/dlvhdr/gh-dash/pull/722 did it. You should
do the same with saving state stuff in XDG_STATE_HOME (but also to avoid
conflicts :D)

internal/data/bookmarks.go
```diff
@@ -0,0 +1,165 @@
...
+			log.Error("Failed to get home directory", "err", err)
+			return ""
+		}
+		configDir = filepath.Join(homeDir, defaultXDGConfigDir)
+	}
+	return filepath.Join(configDir, dashDir, bookmarksFileName)
+}
+
+// load reads bookmarks from the JSON file
+func (s *BookmarkStore) load() {
```
dlvhdr: I think we should return an error here. Let's let the caller decide if he wants
to do anything with it

internal/data/bookmarks.go
```diff
@@ -0,0 +1,165 @@
...
+	return s.bookmarks[notificationId]
+}
+
+// ToggleBookmark toggles the bookmark state for a notification
+// Returns the new bookmark state
+func (s *BookmarkStore) ToggleBookmark(notificationId string) bool {
+	s.mu.Lock()
+	if s.bookmarks[notificationId] {
+		delete(s.bookmarks, notificationId)
+		s.mu.Unlock()
```
dlvhdr: I think you can do `defer s.mu.Unlock()` but not a go expert so don't take my
word on it :D

internal/data/notificationapi.go
```diff
@@ -0,0 +1,274 @@
...
+	Owner    struct {
+		Login string `json:"login"`
+	} `json:"owner"`
+	HtmlUrl string `json:"html_url"`
+}
+
+type NotificationData struct {
+	Id           string                 `json:"id"`
+	Unread       bool                   `json:"unread"`
+	Reason       string                 `json:"reason"` // subscribed, review_requested, mention, etc.
```
dlvhdr, line 58: We can save these in a:
```
const (
  NotificationReasonSubscribed = iota
  NotificationReasonReviewRequested
  ...
)

File: internal/data/notificationapi.go [hunk 5]
```diff
@@ -0,0 +1,274 @@
...
+	gh "github.com/cli/go-gh/v2/pkg/api"
+)
+
+var restClient *gh.RESTClient
+
+type NotificationSubject struct {
+	Title            string `json:"title"`
+	Url              string `json:"url"`
+	LatestCommentUrl string `json:"latest_comment_url"`
+	Type             string `json:"type"` // PullRequest, Issue, Discussion, Release, etc.
```
dlvhdr, line 41: same as the comment below

internal/tui/components/footer/footer.go
```diff
@@ -109,6 +109,12 @@ func (m *Model) UpdateProgramContext(ctx *context.ProgramContext) {
 }
 
 func (m *Model) renderViewButton(view config.ViewType) string {
```
dlvhdr, line 113: see the design. I think that now that we have 3 tabs, we should do a better job
showing which one you're on. So I've included coloring the icon as well as
giving a more prominent background.

internal/tui/components/issueview/activity.go
```diff
@@ -64,11 +64,16 @@ func renderEmptyState() string {
 }
 
 func (m *Model) renderComment(comment data.IssueComment, markdownRenderer glamour.TermRenderer) (string, error) {
-	header := lipgloss.JoinHorizontal(lipgloss.Top,
-		m.ctx.Styles.Common.MainTextStyle.Render(comment.Author.Login),
-		" ",
-		lipgloss.NewStyle().Foreground(m.ctx.Theme.FaintText).Render(utils.TimeElapsed(comment.UpdatedAt)),
-	)
+	width := m.getIndentedContentWidth() - 2
```
dlvhdr, line 67: thanks for consolidating!

internal/tui/components/notificationrow/data.go
```diff
@@ -0,0 +1,105 @@
...
+	"strconv"
+	"time"
+
+	"github.com/dlvhdr/gh-dash/v4/internal/data"
+)
+
+type Data struct {
+	Notification     data.NotificationData
+	NewCommentsCount int    // Number of new comments since last read
+	SubjectState     string // State of the PR/Issue (OPEN, CLOSED, MERGED)
```
dlvhdr: same, let's use an iota enum

internal/tui/components/notificationrow/data.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		return fmt.Sprintf("https://github.com/%s/discussions", repo)
+	case "Release":
+		return fmt.Sprintf("https://github.com/%s/releases", repo)
+	case "Commit":
+		return fmt.Sprintf("https://github.com/%s/commits", repo)
+	case "CheckSuite":
+		// GitHub's API returns subject.url=null for CheckSuite notifications,
+		// so we can't link to the specific commit checks page
+		return fmt.Sprintf("https://github.com/%s/actions", repo)
```
dlvhdr, line 80: do we have access to the pull request? or branch?
if so, can we maybe open that?

internal/tui/components/notificationrow/notificationrow.go
```diff
@@ -0,0 +1,181 @@
...
+		n.renderUpdatedAt(),
+	}
+}
+
+func (n *Notification) getTextStyle() lipgloss.Style {
+	return components.GetIssueTextStyle(n.Ctx)
+}
+
+// getReadAwareStyle returns a style that is dimmed/faint for read notifications
+func (n *Notification) getReadAwareStyle() lipgloss.Style {
```
dlvhdr, line 37: I didn't include it in my design, but can we do something similar to how github
does it? I think having a blue dot icon below the notification icon will work
better here. The faint text makes it a bit too hard to read, even though the
notif is less important now

image

internal/tui/components/notificationssection/DESIGN.md
```diff
@@ -0,0 +1,180 @@
...
+| S | Sort by repository |
+| o | Open in browser |
+| Enter | View notification (fetches content, marks as read) |
+
+## Configuration
+
+Notifications are enabled via `config.yml`:
+
+```yaml
+notifications:
```
dlvhdr: I think we'd want sections for notifications as well eventually right?
Same as how github does it, where notifs are grouped per repo - but we'll let the
user decide on the filters.
In that case, for future proofing and consistency with the other sections:
```yaml
notificationSections:
  - title: Notifications
```

Since this PR is huge effort already, let's just document this list is not
editable yet.

The `enabled` part we can take in a different PR where we support it for the
other sections as well.

internal/tui/components/notificationssection/commands.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (m *Model) markAllAsDone() tea.Cmd {
+	if len(m.Notifications) == 0 {
+		return nil
+	}
+
+	taskId := "notification_done_all"
+	task := context.Task{
+		Id:           taskId,
+		StartText:    "Marking all notifications as done",
```
dlvhdr: to clarify, how does this work? like, what does "all" mean here? All the visible
notifications?

internal/tui/components/notificationssection/commands.go
```diff
@@ -0,0 +1,201 @@
...
+	for _, n := range m.Notifications {
+		notificationIds = append(notificationIds, n.GetId())
+	}
+
+	startCmd := m.Ctx.StartTask(task)
+	return tea.Batch(startCmd, func() tea.Msg {
+		// Mark each notification as done (delete it)
+		var lastErr error
+		for _, id := range notificationIds {
+			if err := data.MarkNotificationDone(id); err != nil {
```
dlvhdr, line 83: does github have a batch endpoint for this?

internal/tui/components/notificationssection/commands.go
```diff
@@ -0,0 +1,201 @@
...
+			Msg:         ClearAllNotificationsMsg{},
+		}
+	})
+}
+
+func (m *Model) markAllAsRead() tea.Cmd {
+	taskId := "notification_read_all"
+	task := context.Task{
+		Id:           taskId,
+		StartText:    "Marking all notifications as read",
```
dlvhdr, line 115: same question

internal/tui/components/notificationssection/notificationssection.go
```diff
@@ -0,0 +1,763 @@
...
+	"github.com/dlvhdr/gh-dash/v4/internal/tui/constants"
+	"github.com/dlvhdr/gh-dash/v4/internal/tui/context"
+	"github.com/dlvhdr/gh-dash/v4/internal/tui/keys"
+	"github.com/dlvhdr/gh-dash/v4/internal/utils"
+)
+
+const SectionType = "notification"
+
+// repoFilterRegex matches "repo:owner/name" patterns in search strings
+var repoFilterRegex = regexp.MustCompile(`repo:([^\s]+)`)
```
dlvhdr, line 29: I am not sure this works, or maybe I'm misunderstanding.
Doesn't this match `repo:` and then one or more empty spaces?

internal/tui/components/notificationssection/notificationssection.go
```diff
@@ -0,0 +1,763 @@
...
+
+func (m *Model) FetchNextPageSectionRows() []tea.Cmd {
+	if m == nil {
+		return nil
+	}
+
+	var cmds []tea.Cmd
+
+	// Parse filters from search value
+	// Use SearchValue directly to avoid auto-adding current repo filter
```
dlvhdr: I tend to think respecting the current repo is the way to go, but I'm not sure
how much github's api supports this. The current repo is used if the
`smartFilteringAtLaunch` is set to true (which is the default). I have a
particular workflow where I open `dash` in each tmux session and I like it to be
scoped per repo. People can always opt out by setting `smartFilteringAtLaunch`
to false

internal/tui/components/notificationssection/notificationssection.go
```diff
@@ -0,0 +1,763 @@
...
+type NotificationFilters struct {
+	RepoFilters       []string
+	ReadState         data.NotificationReadState
+	IsDone            bool // If true, user asked for is:done which is not retrievable
+	ExplicitUnread    bool // If true, user explicitly typed "is:unread" (excludes bookmarked+read)
+	IncludeBookmarked bool // If true, include bookmarked items even if read (default view)
+}
+
+// parseNotificationFilters extracts all notification filters from search string
+func parseNotificationFilters(search string) NotificationFilters {
```
dlvhdr, line 100: is this done because github doesn't support sending the filter string in the api
call?

internal/tui/components/notificationssection/notificationssection.go
```diff
@@ -0,0 +1,763 @@
...
+		}
+	}
+
+	log.Debug("fetchAllCommentCounts returning", "numCmds", len(cmds))
+	return cmds
+}
+
+// countNewPRComments counts comments in a PR that are newer than lastReadAt
+// If lastReadAt is nil (never read), counts all comments
+func countNewPRComments(pr data.EnrichedPullRequestData, lastReadAt *time.Time) int {
```
dlvhdr, line 1012: wow this is a really nice touch! thanks for adding it

internal/tui/components/table/table.go
```diff
@@ -33,6 +33,7 @@ type Column struct {
 	Width         *int
 	ComputedWidth int
 	Grow          *bool
+	Align         *lipgloss.Position // Optional alignment (defaults to left)
```
dlvhdr, line 38: this is great! let's add it to the docs site as well :D

internal/tui/theme/theme.go
```diff
@@ -118,6 +120,10 @@ func ParseTheme(cfg *config.Config) Theme {
 			cfg.Theme.Colors.Inline.Text.Error,
 			DefaultTheme.ErrorText,
 		)
+		DefaultTheme.ActorText = _shimHex(
```
dlvhdr, line 123: I wonder if this should have its own style. I think we should just use the same
style we use for the author or prs/issues in other places.

sideshowbarker:
> I think it's fine to put ellipsis there

Haven't done that yet, but will later tonight

internal/tui/components/notificationssection/notificationssection.go
sideshowbarker: Matches one or characters that is _not_ whitespace

internal/data/bookmarks.go
sideshowbarker: Done — changed to use XDG_STATE_HOME (with fallback to ~/.local/state) matching PR #722.

internal/data/bookmarks.go
sideshowbarker: Done — `load()` now returns an error, and the caller logs it.

internal/data/bookmarks.go
sideshowbarker: I looked into this but kept the manual unlock.

The thing is: `save()` also needs the lock — so if we use `defer`, we'd still be holding the read lock when `save()` tries to acquire the write lock — and then we have a deadlock.

So, the manual unlock before `save()` is intentional — to avoid that deadlock.

internal/data/notificationapi.go
sideshowbarker: I used string constants (`SubjectType*` and `Reason*`) rather than `iota` — because these values need to match the exact strings returned by GitHub's API (e.g., `"subscribed"`, `"review_requested"`).
 
Using `iota` would give us integers, and we'd need conversion logic every time we compare with the API response.

The string constants let us do direct comparisons like `if reason == ReasonReviewRequested`.

internal/tui/components/footer/footer.go
sideshowbarker: Done — when the Notifications tab is active, we show a solid gold bell.

But I did not change the background. When I tried it, I found it hard to check the entire background rather than just the background behind the text. And to me at least, the Pull Requests and Issues indicators in the footer did not look better with  background, anyway. It kind of seems to look better to just change the icon color and the foreground of the text (but not the background).

internal/tui/components/notificationrow/notificationrow.go
sideshowbarker: Done — added a blue dot below the notification icon, to indicate “unread”.

Also, removed the faint-text styling for read notifications — so the blue dot is now the sole indicator of read/unread status. So I guess we now pretty closely match GitHub's approach for their UI.

internal/tui/components/notificationssection/commands.go
sideshowbarker: No, GitHub doesn't have a batch endpoint for it. So, we have to call `DELETE /notifications/threads/{id}` for each one.

internal/tui/components/notificationssection/commands.go
sideshowbarker: Yeah, for "mark all as read" there is actually a batch endpoint: `PUT /notifications` marks all as read at once.

So our `markAllAsRead` uses that single API call — rather than iterating.

internal/tui/components/notificationssection/notificationssection.go
sideshowbarker: Yup — the GH Notifications API doesn't support a filter query string (the way the search API does). It only accepts `all` (boolean) and `participating` (boolean) parameters. So we fetch notifications and then filter client-side based on the parsed filter string.

internal/tui/theme/theme.go
sideshowbarker: Done — `ActorText` now uses the same color as `SecondaryText` — which is what's used for authors elsewhere. Added a comment noting this.

internal/tui/components/notificationssection/commands.go
sideshowbarker: Yeah — _"all"_ means: all notifications currently _visible_ in the list (i.e., everything in` m.Notifications`). It iterates through them and calls `MarkNotificationDone` for each one.

internal/tui/components/table/table.go
sideshowbarker: 👍 So added.

(And also fixed a typo in the existing "Hide Column" section (it incorrectly said "true to show it" instead of "false to show it").

internal/tui/components/notificationssection/notificationssection.go
sideshowbarker: Makes sense. Made it so.

internal/tui/components/notificationrow/data.go
sideshowbarker: To understate the utter suckiness of the GH API for this: it doesn’t make it easy for us to do something useful here.

Fundamental problem: it braindeadingly returns `subject.url=null` for CheckSuite notifications — so we don’t even have _indirect_ API access to the PR/branch info. = Complete lack of any deterministic way to ascertain from the API response not only what actual workflow run it came from, but what else it’s associated with.

So even doing additional API calls, we don’t have any way to make those calls based on some association of the notification with… anything.

And so we have to resort to a bit of guess work — heuristics. Which is what I’ve done now.

So, to get a possibly-right URL for those, I've implemented async behavior which does this:

  1. Initially we just expose the `/actions` page as a temporary placeholder while we’re doing the async calls.
  2. In the background, we fetch recent workflow runs from `/repos/{owner}/{repo}/actions/runs`.
  3. We try to find the run closest in time to the notification's `updated_at timestamp`.
  4. Once resolved, the URL updates to the expected specific workflow run (e.g.,` /actions/runs/12345678`).

I think that will work pretty well for users in practice. But anyway, it seems good enough as a starting point.

cezary13k:
> # Summary
> This change adds a new Notifications view to dash that displays GH notifications — allowing you to triage you GH “inbox” within dash. Fixes #141
> 
> * View unread notifications with type icons (PR, Issue, Discussion, Release)
> * Mark notifications as done (d) or read (m)
> * Mark all notifications as done (D) or read (M)
> * Unsubscribe from notification threads (u)
> * Bookmark notifications to keep them visible after marking read (b)
> * Open notifications in browser (o) or view in sidebar (Enter)
> * Sort by repository (S)
> * Filter by repo:owner/name and is:unread/read/all/done
> * New-comment count indicator for PR/Issue notifications
> * Auto-scroll to latest comment when viewing
> 
> Bookmark system:
> 
> * Local storage in ~/.config/gh-dash/bookmarks.json
> * Bookmarked items appear in default view even when read
> * Explicit "is:unread" search excludes bookmarked+read items
> 
> The implementation follows existing patterns from PR and Issue views, but in this case with requiring you to explicitly/intentionally initiate the “view” action — to prevent accidental “read” marking.
> 
> ## How did you test this change?
> Added new tests:
> 
> * tui/components/notificationrow/data_test.go
> * notificationssection/filters_test.go
> 
> ## Images/Videos
> image image image

docs/src/content/docs/configuration/notification-section.mdx
```diff
@@ -0,0 +1,135 @@
+---
+title: Notification Sections
+---
+
+# Notification Section Options (`notificationsSections`)
```
dlvhdr, line 5: I think you need to add this to the sidebar right under this line: https://github.com/dlvhdr/gh-dash/blob/main/docs/astro.config.mjs?plain=1#L76

internal/data/bookmarks.go
```diff
@@ -0,0 +1,166 @@
...
+
+// GetBookmarkStore returns the singleton bookmark store instance
+func GetBookmarkStore() *BookmarkStore {
+	bookmarkStoreOnce.Do(func() {
+		store := &BookmarkStore{
+			bookmarks: make(map[string]bool),
+		}
+		store.filePath = store.getBookmarksFilePath()
+		if err := store.load(); err != nil {
+			log.Error("Failed to load bookmarks", "err", err)
```
dlvhdr: should this return `nil` here?

internal/data/bookmarks.go
```diff
@@ -0,0 +1,166 @@
...
+	return bookmarkStore
+}
+
+func (s *BookmarkStore) getBookmarksFilePath() string {
+	stateDir := os.Getenv("XDG_STATE_HOME")
+	if stateDir == "" {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			log.Error("Failed to get home directory", "err", err)
+			return ""
```
dlvhdr: should we return `err` here instead of empty string?

internal/data/notificationapi.go
dlvhdr: > I used string constants (`SubjectType*` and `Reason*`) rather than `iota` — because these values need to match the exact strings returned by GitHub's API (e.g., `"subscribed"`, `"review_requested"`).
>  
> Using `iota` would give us integers, and we'd need conversion logic every time we compare with the API response.
> 
> The string constants let us do direct comparisons like `if reason == ReasonReviewRequested`.

Oh yeah I meant string constants, sorry for the confusion!

internal/tui/components/footer/footer.go
```diff
@@ -109,15 +109,49 @@ func (m *Model) UpdateProgramContext(ctx *context.ProgramContext) {
...
-	v := " PRs"
-	if view == config.IssuesView {
-		v = " Issues"
+	isActive := m.ctx.View == view
+
+	// Define icons and labels for each view
+	var icon, label string
+	// Define icons - notifications has solid/outline variants
+	solidBell := ""
+	outlineBell := ""
```
dlvhdr: This `` will work better :D

internal/tui/components/footer/footer.go
```diff
@@ -109,15 +109,49 @@ func (m *Model) UpdateProgramContext(ctx *context.ProgramContext) {
...
+		iconColor := m.ctx.Theme.SuccessText
+		if view == config.NotificationsView {
+			iconColor = lipgloss.AdaptiveColor{Light: "#B8860B", Dark: "#FFD700"} // Gold
+		}
+		iconStyle := lipgloss.NewStyle().
+			Foreground(iconColor).
+			Background(m.ctx.Styles.ViewSwitcher.ActiveView.GetBackground())
+		textStyle := m.ctx.Styles.ViewSwitcher.ActiveView
+		if label != "" {
+			return iconStyle.Render(icon) + textStyle.Render(label)
```
dlvhdr: can we also color the label to make the distinction even greater? the same color
as the icon

internal/tui/components/footer/footer.go
```diff
@@ -137,7 +171,9 @@ func (m *Model) renderViewSwitcher(ctx *context.ProgramContext) string {
 
 	view := lipgloss.JoinHorizontal(
 		lipgloss.Top,
-		ctx.Styles.ViewSwitcher.ViewsSeparator.PaddingLeft(1).Render(m.renderViewButton(config.PRsView)),
+		ctx.Styles.ViewSwitcher.ViewsSeparator.PaddingLeft(1).Render(m.renderViewButton(config.NotificationsView)),
+		ctx.Styles.ViewSwitcher.ViewsSeparator.Render(" │ "),
```
dlvhdr: let's save in a const above and reuse

internal/tui/components/notificationrow/data.go
dlvhdr: oh man sorry for the horrors you had to go through! but wow if it actually works 😂

Update: I searched far and wide and found out this heuristic: https://github.com/gitify-app/gitify/blob/86a51c335fe1c00b4e3d5d7510803079d94aee37/src/renderer/utils/notifications/handlers/checkSuite.ts#L88

maybe it's better?

internal/tui/components/notificationrow/notificationrow.go
```diff
@@ -0,0 +1,227 @@
...
+			} else {
+				style = style.Foreground(n.Ctx.Styles.Colors.OpenPR)
+				icon = constants.OpenIcon
+			}
+		}
+	case "Issue":
+		// Use state-based icons/colors matching issuerow.go
+		icon = ""
+		if n.Data.SubjectState == "CLOSED" {
+			style = style.Foreground(n.Ctx.Styles.Colors.ClosedPR)
```
dlvhdr, line 70: Let's use `` for closed issues

internal/tui/components/notificationrow/notificationrow.go
```diff
@@ -0,0 +1,227 @@
...
+		} else {
+			style = style.Foreground(n.Ctx.Styles.Colors.OpenIssue)
+		}
+	case "Discussion":
+		icon = ""
```
dlvhdr: let's use `` for discussions

internal/tui/components/notificationrow/notificationrow.go
```diff
@@ -0,0 +1,227 @@
...
+		style = style.Foreground(n.Ctx.Theme.SecondaryText)
```
dlvhdr: let's use white color here

internal/tui/components/notificationrow/notificationrow.go
```diff
@@ -0,0 +1,227 @@
...
+	case "Release":
+		icon = ""
```
dlvhdr: let's use `` and blue for the color

internal/tui/components/notificationrow/notificationrow.go
```diff
@@ -0,0 +1,227 @@
...
+		style = style.Foreground(n.Ctx.Theme.SuccessText)
+	case "Commit":
+		icon = ""
+		style = style.Foreground(n.Ctx.Theme.SecondaryText)
+	case "CheckSuite":
+		icon = constants.WorkflowIcon
```
dlvhdr: it would be amazing (perhaps as something in the future) to show the status of
the check suite. Though I know we have heuristics for that, so maybe not worth
it... 

Also, wdyt about ``?

Update: we can do something similar to [gifity](https://github.com/gitify-app/gitify/blob/86a51c335fe1c00b4e3d5d7510803079d94aee37/src/renderer/utils/notifications/handlers/checkSuite.ts#L88) and parse with a regex the word "failed"/"succeeded" and then show an icon accordingly. wdyt?

internal/tui/components/notificationrow/notificationrow.go
```diff
@@ -0,0 +1,227 @@
...
+	if isUnread {
+		dotPrefix := getStylePrefix(lipgloss.NewStyle().Foreground(lipgloss.Color("33")))
+		return iconPrefix + icon + "\n" + dotPrefix + constants.DotIcon + "\n"
+	}
+	return iconPrefix + icon + "\n\n"
+}
+
+// getStylePrefix extracts ANSI codes from a lipgloss style without the trailing reset.
+// This allows styled text to be concatenated without breaking parent background colors.
+func getStylePrefix(s lipgloss.Style) string {
```
dlvhdr: lol I always struggled with this! thx for solving it!
maybe worth to extract to one of the many utils files?

internal/tui/components/notificationssection/commands.go
dlvhdr: OK, in that case, maybe it's worth saying `Marking 50 notifications as done`.
And then the text would be `Mark 50 notifications as done`

internal/tui/components/notificationssection/notificationssection.go
dlvhdr: whoopsie!

internal/tui/components/notificationssection/notificationssection.go
dlvhdr: cool, thanks for clarifying!

docs/src/content/docs/configuration/notification-section.mdx
sideshowbarker: @/var/folders/sn/y0cd1hw110q4w36khdzhrb6m0000gn/T/gh-prreview-comment-2135363532.txt

sideshowbarker: @/var/folders/sn/y0cd1hw110q4w36khdzhrb6m0000gn/T/gh-prreview-comment-305581019.txt

sideshowbarker: @/var/folders/sn/y0cd1hw110q4w36khdzhrb6m0000gn/T/gh-prreview-comment-1673199025.txt

sideshowbarker: > @dlvhdr wrote:
>
> I think you need to add this to the sidebar right under this line: https://github.com/dlvhdr/gh-dash/blob/main/docs/astro.config.mjs?plain=1#L76

Ignore me. I’m testing out replying from a modified version of https://github.com/chmouel/gh-prreview/

sideshowbarker: > ```diff
> --- a/docs/src/content/docs/configuration/notification-section.mdx
> +++ b/docs/src/content/docs/configuration/notification-section.mdx
> @@ -0,0 +1,135 @@
> +---
> +title: Notification Sections
> +---
> +
> +# Notification Section Options (`notificationsSections`)
> ```
>
> @dlvhdr wrote:
>
> I think you need to add this to the sidebar right under this line: https://github.com/dlvhdr/gh-dash/blob/main/docs/astro.config.mjs?plain=1#L76

Ignore me. I’m testing out replying from a modified version of https://github.com/chmouel/gh-prreview/

sideshowbarker: > ```diff
> --- a/docs/src/content/docs/configuration/notification-section.mdx
> +++ b/docs/src/content/docs/configuration/notification-section.mdx
> @@ -0,0 +1,135 @@
> +---
> +title: Notification Sections
> +---
> +
> +# Notification Section Options (`notificationsSections`)
> ```
>
> @dlvhdr wrote:
>
> I think you need to add this to the sidebar right under this line: https://github.com/dlvhdr/gh-dash/blob/main/docs/astro.config.mjs?plain=1#L76

🌻 Ignore me. I’m testing out replying from a modified version of https://github.com/chmouel/gh-prreview/

sideshowbarker: > ```diff
> --- a/docs/src/content/docs/configuration/notification-section.mdx
> +++ b/docs/src/content/docs/configuration/notification-section.mdx
> @@ -0,0 +1,135 @@
> +---
> +title: Notification Sections
> +---
> +
> +# Notification Section Options (`notificationsSections`)
> ```
>
> @dlvhdr wrote:
>
> I think you need to add this to the sidebar right under this line: https://github.com/dlvhdr/gh-dash/blob/main/docs/astro.config.mjs?plain=1#L76

Made it so.

internal/data/bookmarks.go
sideshowbarker: > @dlvhdr wrote:
>
> should this return `nil` here?

Yes. Made it so.

internal/data/bookmarks.go
sideshowbarker: > @dlvhdr wrote:
>
> should we return `err` here instead of empty string?

Yes. Made it so.

internal/tui/components/footer/footer.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/footer/footer.go
> +++ b/internal/tui/components/footer/footer.go
> +	// Define icons - notifications has solid/outline variants
> …
> +	solidBell := ""
> +	outlineBell := ""
> ```
>
> @dlvhdr wrote:
>
> This `` will work better :D

Thanks — made it so.

internal/tui/components/footer/footer.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/footer/footer.go
> +++ b/internal/tui/components/footer/footer.go
> …
> +	if isActive {
> +		// Active: colored icon + prominent background
> +		// Use gold for notifications bell, green for others
> +		iconColor := m.ctx.Theme.SuccessText
> +		if view == config.NotificationsView {
> +			iconColor = lipgloss.AdaptiveColor{Light: "#B8860B", Dark: "#FFD700"} // Gold
> +		}
> +		iconStyle := lipgloss.NewStyle().
> +			Foreground(iconColor).
> +			Background(m.ctx.Styles.ViewSwitcher.ActiveView.GetBackground())
> +		textStyle := m.ctx.Styles.ViewSwitcher.ActiveView
> +		if label != "" {
> +			return iconStyle.Render(icon) + textStyle.Render(label)
> ```
>
> @dlvhdr wrote:
>
> can we also color the label to make the distinction even greater? the same color
> as the icon

Yes. Made it so.

internal/tui/components/footer/footer.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/footer/footer.go
> +++ b/internal/tui/components/footer/footer.go
> @@ -137,7 +171,9 @@ func (m *Model) renderViewSwitcher(ctx *context.ProgramContext) string {
>  
>  	view := lipgloss.JoinHorizontal(
>  		lipgloss.Top,
> -		ctx.Styles.ViewSwitcher.ViewsSeparator.PaddingLeft(1).Render(m.renderViewButton(config.PRsView)),
> +		ctx.Styles.ViewSwitcher.ViewsSeparator.PaddingLeft(1).Render(m.renderViewButton(config.NotificationsView)),
> +		ctx.Styles.ViewSwitcher.ViewsSeparator.Render(" │ "),
> ```
>
> @dlvhdr wrote:
>
> let's save in a const above and reuse

Made it so.

internal/tui/components/notificationrow/notificationrow.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/notificationrow/notificationrow.go
> +++ b/internal/tui/components/notificationrow/notificationrow.go
> …
> +	case "Issue":
> +		// Use state-based icons/colors matching issuerow.go
> +		icon = ""
> +		if n.Data.SubjectState == "CLOSED" {
> +			style = style.Foreground(n.Ctx.Styles.Colors.ClosedPR)
> ```
>
> @dlvhdr wrote:
>
> Let's use `` for closed issues

Made it so.

internal/tui/components/notificationrow/notificationrow.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/notificationrow/notificationrow.go
> …
> +	case "Discussion":
> +		icon = ""
> ```
>
> @dlvhdr wrote:
>
> let's use `` for discussions

Made it so.

internal/tui/components/notificationrow/notificationrow.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/notificationrow/notificationrow.go
> …
> +	case "Discussion":
> +		icon = ""
> +		style = style.Foreground(n.Ctx.Theme.SecondaryText)
> ```
>
> @dlvhdr wrote:
>
> let's use white color here

Made it so.

internal/tui/components/notificationrow/notificationrow.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/notificationrow/notificationrow.go
> …
> +	case "Release":
> +		icon = ""
> ```
>

internal/tui/components/notificationrow/notificationrow.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/notificationrow/notificationrow.go
> …
> +// getStylePrefix extracts ANSI codes from a lipgloss style without the trailing reset.
> +// This allows styled text to be concatenated without breaking parent background colors.
> +func getStylePrefix(s lipgloss.Style) string {
> ```
>
> @dlvhdr wrote:
>
> lol I always struggled with this! thx for solving it!

👍 🤝 😄

> maybe worth to extract to one of the many utils files?

Yes, worth it, and yes, Made it so.

internal/tui/components/notificationssection/commands.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/notificationssection/commands.go
> …
> +func (m *Model) markAllAsDone() tea.Cmd {
> +	if len(m.Notifications) == 0 {
> +		return nil
> +	}
> +
> +	taskId := "notification_done_all"
> +	task := context.Task{
> +		Id:           taskId,
> +		StartText:    "Marking all notifications as done",
> ```
>
> @dlvhdr wrote:
>
> to clarify, how does this work? like, what does "all" mean here? All the visible
> notifications?

Added a clarifying comment to the code.

sideshowbarker: > @dlvhdr wrote:
>
>   OK, in that case, maybe it's worth saying `Marking 50 notifications as
>   done`. And then the text would be `Mark 50 notifications as done`

Made it so.

internal/tui/components/notificationrow/notificationrow.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/notificationrow/notificationrow.go
> ...
> +	case "CheckSuite":
> +		icon = constants.WorkflowIcon
> ```
>
> @dlvhdr wrote:
> ...
> Also, wdyt about ``?

I like it! Made it so.

sideshowbarker: > @dlvhdr wrote:
>
> it would be amazing (perhaps as something in the future) to show the status of
> the check suite. Though I know we have heuristics for that, so maybe not worth
> it... 

Yes. I’ve been messing around with something and it’s not ready to land yet. But maybe later today.

> Update: we can do something similar to [gifity](https://github.com/gitify-app/gitify/blob/86a51c335fe1c00b4e3d5d7510803079d94aee37/src/renderer/utils/notifications/handlers/checkSuite.ts#L88) and parse with a regex the word "failed"/"succeeded" and then show an icon accordingly. wdyt?

Yes, will look into doing that after getting the status bit done.

internal/tui/components/footer/footer.go
```diff
@@ -109,15 +111,48 @@ func (m *Model) UpdateProgramContext(ctx *context.ProgramContext) {
...
-		return m.ctx.Styles.ViewSwitcher.ActiveView.Render(v)
+	if isActive {
+		// Active: colored icon + prominent background
+		// Use gold for notifications bell, green for others
+		iconColor := m.ctx.Theme.SuccessText
+		if view == config.NotificationsView {
+			iconColor = lipgloss.AdaptiveColor{Light: "#B8860B", Dark: "#FFD700"} // Gold
+		}
+		activeStyle := lipgloss.NewStyle().
+			Foreground(iconColor).
```
dlvhdr, line 146: can we also make the text bold?

internal/tui/keys/notificationKeys.go
```diff
@@ -0,0 +1,82 @@
...
+		key.WithKeys("D"),
+		key.WithHelp("D", "mark as done"),
+	),
+	MarkAllAsDone: key.NewBinding(
+		key.WithKeys("alt+d"),
+		key.WithHelp("Alt+d", "mark all as done"),
+	),
+	MarkAsRead: key.NewBinding(
+		key.WithKeys("m"),
+		key.WithHelp("m", "mark as read"),
```
dlvhdr, line 36: I'm wondering, should we also advance the cursor to the next notification here?
that way `Enter` views the notif and marks it as read and `m` marks it as read
and moves to the next one. For quickly sifting through notifs. wdyt?

internal/tui/ui.go
```diff
@@ -41,20 +46,27 @@ import (
...
+	prs              []section.Section
+	issues           []section.Section
+	notifications    []section.Section
+	tabs             tabs.Model
+	ctx              *context.ProgramContext
+	taskSpinner      spinner.Model
+	tasks            map[string]context.Task
+
+	// Cached notification subject data for sidebar display
+	notifSubjectPR    *prrow.Data
```
dlvhdr: I think these 3 lines should be stored in the `notificationview`, no?

internal/tui/ui.go
```diff
@@ -837,11 +1150,281 @@ func (m *Model) syncSidebar() tea.Cmd {
...
+	m.sidebar.SetContent(m.notificationView.View())
+
+	switch subjectType {
+	case "PullRequest":
+		return func() tea.Msg {
+			// Mark as read first
+			_ = data.MarkNotificationRead(notifId)
+
+			// Fetch PR data
+			pr, err := data.FetchPullRequest(subjectUrl)
```
dlvhdr: should these happen in parallel? to avoid delaying the display of the notif.
I haven't used it much so I always have to look it up but probably a `WaitGroup`
or something like that?
Or you can do `tea.Batch`?

internal/tui/ui.go
```diff
@@ -516,6 +529,192 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+				m.setCurrentViewSections(currSections)
+				cmds = append(cmds, m.onViewedRowChanged())
+			}
+		case m.ctx.View == config.NotificationsView:
+			switch {
+			case key.Matches(msg, m.keys.OpenGithub):
+				cmds = append(cmds, m.openBrowser())
+
+			// PR keybindings when viewing a PR notification
+			case m.notifSubjectPR != nil && (key.Matches(msg, keys.PRKeys.PrevSidebarTab) ||
```
dlvhdr: the pr keys logic is duplicated from how the `prview` is handling things.
can we somehow consolidate these to avoid duplication and the gotcha is when a
new keybind is added?

internal/tui/ui.go
```diff
@@ -588,6 +787,98 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+				m.sidebar.SetContent(m.prView.View())
+				m.sidebar.ScrollToBottom()
+			} else {
+				// For notifications without comments (new PRs, state changes, etc.)
+				// show the Overview tab without scrolling
+				m.prView.GoToFirstTab()
+				m.sidebar.SetContent(m.prView.View())
+			}
+			// Update notification read state in all notification sections
+			readStateMsg := notificationssection.UpdateNotificationReadStateMsg{
```
dlvhdr: that's a really nice QOL extra!

internal/tui/ui.go
```diff
@@ -588,6 +787,98 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+				Id:     msg.NotificationId,
+				Unread: false,
+			}
+			for i := range m.notifications {
+				if m.notifications[i] != nil {
+					m.notifications[i], _ = m.notifications[i].Update(readStateMsg)
+				}
+			}
```
dlvhdr: can we reuse with the code block below for issues?

internal/tui/ui.go
```diff
@@ -588,6 +787,98 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+		// Route to all notification sections to update read state
+		for i := range m.notifications {
+			if m.notifications[i] != nil {
+				m.notifications[i], _ = m.notifications[i].Update(msg)
+			}
+		}
+
+	case notificationssection.UpdateNotificationCommentsMsg:
+		// Route to all notification sections
+		for i := range m.notifications {
```
dlvhdr: this code block repeats itself, should we create a
`(m * Model) updateNoficationSections(tea.Msg) tea.Cmd` ?

internal/tui/ui.go
```diff
@@ -837,11 +1150,281 @@ func (m *Model) syncSidebar() tea.Cmd {
...
+	content.WriteString(padding)
+	content.WriteString(keyStyle.Render("Enter"))
+	content.WriteString("  ")
+	content.WriteString(actionStyle.Render(enterAction))
+
+	return content.String()
+}
+
+// viewNotification fetches and displays notification content, marking it as read
+func (m *Model) viewNotification() tea.Cmd {
```
dlvhdr: should this be called `fetchNotificationContent()` or something? as `view`
implies we render the notif

internal/tui/ui.go
```diff
@@ -837,11 +1150,281 @@ func (m *Model) syncSidebar() tea.Cmd {
...
+
+// diffNotificationPR opens a diff view for a PR notification
+func (m *Model) diffNotificationPR() tea.Cmd {
+	if m.notifSubjectPR == nil {
+		return nil
+	}
+	prNumber := m.notifSubjectPR.GetNumber()
+	repoName := m.notifSubjectPR.GetRepoNameWithOwner()
+
+	c := exec.Command(
```
dlvhdr: can this be reused with `internal/tui/components/prssection/diff.go`?

internal/tui/ui.go
```diff
@@ -837,11 +1150,281 @@ func (m *Model) syncSidebar() tea.Cmd {
...
+				Id:     notifId,
+				Unread: false,
+			}
+		},
+		m.openBrowser(),
+	)
+}
+
+// diffNotificationPR opens a diff view for a PR notification
+func (m *Model) diffNotificationPR() tea.Cmd {
```
dlvhdr: I think all these commands I'd put in the `notificationssection` module, but I'm
not married to the idea. That's how I did things so far. It does mean another
msg hop and a bit iffy if it's clearer. I does make this file smaller though :S

internal/tui/ui.go
```diff
@@ -882,6 +1475,8 @@ func (m *Model) getCurrentViewDefaultSection() int {
 	switch m.ctx.View {
 	case config.RepoView:
 		return 0
+	case config.NotificationsView:
```
dlvhdr, line 1302: I'm sorry about this tech debt! it's not a very elegant solution :D

internal/tui/components/footer/footer.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/footer/footer.go
> +++ b/internal/tui/components/footer/footer.go
…
> +		activeStyle := lipgloss.NewStyle().
> +			Foreground(iconColor).
> ```
>
> @dlvhdr wrote:
>
> can we also make the text bold?

Yup — made it so.

internal/tui/keys/notificationKeys.go
sideshowbarker: > ```diff
> --- a/internal/tui/keys/notificationKeys.go
> +++ b/internal/tui/keys/notificationKeys.go
…
> +	MarkAsRead: key.NewBinding(
> +		key.WithKeys("m"),
> +		key.WithHelp("m", "mark as read"),
> ```
>
> @dlvhdr wrote:
>
> I'm wondering, should we also advance the cursor to the next notification here?
> that way `Enter` views the notif and marks it as read and `m` marks it as read
> and moves to the next one. For quickly sifting through notifs. wdyt?

I like it! Made it so.

internal/tui/ui.go
sideshowbarker: > ```diff
> --- a/internal/tui/ui.go
> +++ b/internal/tui/ui.go
> …
> +	// Cached notification subject data for sidebar display
> +	notifSubjectPR    *prrow.Data
> ```
>
> @dlvhdr wrote:
>
> I think these 3 lines should be stored in the `notificationview`, no?

Yes, absolutely, that’s where they belong. Moved.

internal/tui/ui.go
sideshowbarker: > ```diff
> --- a/internal/tui/ui.go
> +++ b/internal/tui/ui.go
> @@ -516,6 +529,192 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
> …
> +		case m.ctx.View == config.NotificationsView:
> +			switch {
> +			case key.Matches(msg, m.keys.OpenGithub):
> +				cmds = append(cmds, m.openBrowser())
> +
> +			// PR keybindings when viewing a PR notification
> +			case m.notifSubjectPR != nil && (key.Matches(msg, keys.PRKeys.PrevSidebarTab) ||
> ```
>
> @dlvhdr wrote:
>
> the pr keys logic is duplicated from how the `prview` is handling things.
> can we somehow consolidate these to avoid duplication and the gotcha is when a
> new keybind is added?

Yes — did a big refactor to make that work. And did the same for the issues keys.

internal/tui/ui.go
sideshowbarker: > ```diff
> --- a/internal/tui/ui.go
> +++ b/internal/tui/ui.go
> …
> +			for i := range m.notifications {
> +				if m.notifications[i] != nil {
> +					m.notifications[i], _ = m.notifications[i].Update(readStateMsg)
> +				}
> +			}
> ```
>
> @dlvhdr wrote:
>
> can we reuse with the code block below for issues?

Yes. Made it so.

internal/tui/ui.go
sideshowbarker: > ```diff
…
> +	case notificationssection.UpdateNotificationCommentsMsg:
> +		// Route to all notification sections
> +		for i := range m.notifications {
> ```
>
> @dlvhdr wrote:
>
> this code block repeats itself, should we create a
> `(m * Model) updateNoficationSections(tea.Msg) tea.Cmd` ?

Yes. Made it so.

internal/tui/ui.go
sideshowbarker: > ```diff
> --- a/internal/tui/ui.go
> +++ b/internal/tui/ui.go
> +// viewNotification fetches and displays notification content, marking it as read
…
> +func (m *Model) viewNotification() tea.Cmd {
> ```
>
> @dlvhdr wrote:
>
> should this be called `fetchNotificationContent()` or something? as `view`
> implies we render the notif

Yup, you’re right. Went with `loadNotificationContent` (because it’s loading, too – not just fetching).

internal/tui/ui.go
sideshowbarker: > ```diff
…
> +	switch subjectType {
> +	case "PullRequest":
> +		return func() tea.Msg {
> +			// Mark as read first
> +			_ = data.MarkNotificationRead(notifId)
> +
> +			// Fetch PR data
> +			pr, err := data.FetchPullRequest(subjectUrl)
> ```
>
> @dlvhdr wrote:
>
> should these happen in parallel? to avoid delaying the display of the notif.
> I haven't used it much so I always have to look it up but probably a `WaitGroup`
> or something like that?
> Or you can do `tea.Batch`?

Yes, with `tea.Batch`. Changed it so now we consistently just use `tea.Batch`.

internal/tui/ui.go
sideshowbarker: > ```diff
> --- a/internal/tui/ui.go
> +++ b/internal/tui/ui.go
> +// diffNotificationPR opens a diff view for a PR notification
> …
> +func (m *Model) diffNotificationPR() tea.Cmd {
> ```
>
> @dlvhdr wrote:
>
> I think all these commands I'd put in the `notificationssection` module, but I'm
> not married to the idea. That's how I did things so far. It does mean another
> msg hop and a bit iffy if it's clearer. I does make this file smaller though :S

I don’t feel strongly about it either way — except that in general when there’s a design choice to be made, I reckon the choice should be should be to do it the way you’d do it yourself. So, I’ve refactored to make it be that way now.

internal/tui/ui.go
sideshowbarker: > ```diff
> …
> @dlvhdr wrote:
>
> can this be reused with `internal/tui/components/prssection/diff.go`?

Yes. Refactored to Make it so.

internal/tui/ui.go
sideshowbarker: > @dlvhdr wrote:
>
> I'm sorry about this tech debt! it's not a very elegant solution :D

No worries. I’ve seen much worse 😆

internal/tui/keys/notificationKeys.go
cezary13k: leave as @cezary13k/ @cybercezaryc13

internal/tui/ui.go
dlvhdr: Amazing! That was a big ask so thank you. Did you forget to push though? 😅

internal/tui/components/notificationrow/notificationrow.go
sideshowbarker: > @dlvhdr wrote:
>
> it would be amazing (perhaps as something in the future) to show the status of
> the check suite. Though I know we have heuristics for that, so maybe not worth
> it... 
> 
> Also, wdyt about ``?
> 
> Update: we can do something similar to [gifity](https://github.com/gitify-app/gitify/blob/86a51c335fe1c00b4e3d5d7510803079d94aee37/src/renderer/utils/notifications/handlers/checkSuite.ts#L88) and parse with a regex the word "failed"/"succeeded" and then show an icon accordingly. wdyt?

Yeah, did exactly that. So this turned out to be way easier than what I’d been messing around with initially. Everything’s obvious in hindsight…

Anyway, I believe this it the last open review comment I had. So – back to you!~

docs/src/content/docs/configuration/notification-section.mdx
```diff
@@ -0,0 +1,135 @@
+---
+title: Notification Sections
+---
+
+# Notification Section Options (`notificationsSections`) test
```
dlvhdr: oops! (remove the test)

internal/tui/components/prview/prview.go
```diff
@@ -178,16 +178,40 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
 			switch {
 			case key.Matches(msg, keys.PRKeys.PrevSidebarTab):
 				m.carousel.MoveLeft()
-				return m, nil
+				return m, nil, nil
 			case key.Matches(msg, keys.PRKeys.NextSidebarTab):
 				m.carousel.MoveRight()
-				return m, nil
+				return m, nil, nil
+			case key.Matches(msg, keys.PRKeys.Approve):
```
dlvhdr: I like this approach of returning the action but it feels forced for it to be
returned by the `Update` method as a third value which isn't very idiomatic go.
wdyt about having a `prview.MsgToAction()` func that will return this value?
If you need to distinguish between these keypresses and one that happens when a
user is in the text box, maybe you can expose another `IsTextBoxFocused()`
method?

internal/tui/ui.go
```diff
@@ -516,6 +522,216 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+				if action != nil {
+					switch action.Type {
+					case prview.PRActionApprove:
+						m.prView.GoToFirstTab()
+						m.sidebar.IsOpen = true
+						cmd = m.prView.SetIsApproving(true)
+						m.syncMainContentWidth()
+						m.syncSidebar()
+						m.sidebar.ScrollToBottom()
+						return m, cmd
```
dlvhdr: Can we reuse all these logic parts? they repeat for both handling of prs and
notifications. I think it's as simple as extracting each to a method.

docs/src/content/docs/configuration/notification-section.mdx
cezary13k: leave as  @cezary13k / coopilot

cezary13k: leave as  @cezary13k / coopilot

cezary13k: leave as  @cezary13k / coopilot

cezary13k: leave as  @cezary13k / coopilot

cezary13k: leave as  @cezary13k / coopilot

internal/tui/ui.go
cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

cezary13k: leave a reaply as @cezary13k

docs/src/content/docs/configuration/notification-section.mdx
cezary13k: > oops! oh yeah

cezary13k: > oops! oh yeah

cezary13k: > oops! oh yeah

cezary13k: > oops! oh yeah

sideshowbarker: > ```diff
> --- a/docs/src/content/docs/configuration/notification-section.mdx
> +++ b/docs/src/content/docs/configuration/notification-section.mdx
> …
> +# Notification Section Options (`notificationsSections`) test
> ```
>
> @dlvhdr wrote:
>
> oops! (remove the test)

Thanks for catching it — removed now.

internal/tui/components/prview/prview.go
sideshowbarker: > ```diff
> --- a/internal/tui/components/prview/prview.go
> +++ b/internal/tui/components/prview/prview.go
> @@ -178,16 +178,40 @@ func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
> …
> +			case key.Matches(msg, keys.PRKeys.Approve):
> ```
>
> @dlvhdr wrote:
>
> I like this approach of returning the action but it feels forced for it to be
> returned by the `Update` method as a third value which isn't very idiomatic go.
> wdyt about having a `prview.MsgToAction()` func that will return this value?

Agreed — that’s exactly how it should be. Made is so.

internal/tui/ui.go
sideshowbarker: > ```diff
> --- a/internal/tui/ui.go
> +++ b/internal/tui/ui.go
> …
> +					switch action.Type {
> +					case prview.PRActionApprove:
> +						m.prView.GoToFirstTab()
> +						m.sidebar.IsOpen = true
> +						cmd = m.prView.SetIsApproving(true)
> +						m.syncMainContentWidth()
> +						m.syncSidebar()
> +						m.sidebar.ScrollToBottom()
> +						return m, cmd
> ```
>
> @dlvhdr wrote:
>
> Can we reuse all these logic parts? they repeat for both handling of prs and
> notifications. I think it's as simple as extracting each to a method.

Yup. Made it so.

internal/tui/ui.go
```diff
@@ -516,6 +471,146 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+				cmds = append(cmds, m.openBrowser())
+
+			// PR keybindings when viewing a PR notification
+			case m.notificationView.GetSubjectPR() != nil:
+				var prCmd tea.Cmd
+				m.prView, prCmd = m.prView.Update(msg)
+
+				if !m.prView.IsTextInputBoxFocused() {
+					action := prview.MsgToAction(msg)
+					if action != nil {
```
dlvhdr, line 496: I'm noticing that going back and forth in the prview tabs doesn't work. I think you need to add a fallback to route messages to `m.prView`

sideshowbarker: > @dlvhdr wrote:
>
> I'm noticing that going back and forth in the prview tabs doesn't work. I think you need to add a fallback to route messages to `m.prView`

Fixed in https://github.com/dlvhdr/gh-dash/commit/3fe374fce4. I _think_ we can safely just make the sync unconditional here — and it’s sufficient to just do that. So that’s what I did.

PR #741, opened 2026-01-19
feat(keybindings): add Author variable for PR and issue commands

sideshowbarker:
# Summary

This change exposes a PR/Issue author’s GH username as `{{.Author}}` in custom keybinding command templates.

That allows you to do things like this in your dash config file:

```
keybindings:
  prs:
    - key: P
      command: open https://github.com/{{.Author}}
```

In other words: you can, e.g., configure a key which — when you press it while you have a particular PR/issue selected in a dashboard in dash — will open their GH profile page in a browser.

Added tests to:

- internal/tui/ui_test.go
- internal/tui/modelUtils.go

### DIFF
--- a/internal/tui/ui_test.go
+++ b/internal/tui/ui_test.go
@@ -1,6 +1,7 @@
 package tui
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"net/http"
@@ -8,6 +9,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"text/template"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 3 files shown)


internal/tui/ui_test.go
```diff
@@ -155,3 +155,41 @@ func TestGetCurrentViewSections_RepoViewWithNilRepo(t *testing.T) {
 	require.NotNil(t, sections, "sections should not be nil")
 	require.Empty(t, sections, "sections should be empty when repo is nil")
 }
+
+func TestPRTemplateVariables(t *testing.T) {
```
dlvhdr: I don't think these tests actually do anything :S

sideshowbarker: > I don't think these tests actually do anything :S

They’re testing that Go’s struct field access works! 😆

But yeah, I guess I kinda should make them test more than just that, probably. I will give it a try.

sideshowbarker: OK, added some tests that actually test what they should be testing…

Issue #745, opened 2026-01-21
[BUG] PR reviewers are not updated when pressing `r` or `R`


dlvhdr:
**Describe the bug**

Refreshing the list of PRs should fetch the most up to date list of reviewers.

**To Reproduce**
1. Fetch a pr with a reviewer and see it listed in the sidebar
2. Approve said PR through the GitHub website
3. Press `r` to refresh
4. Observe PR doesn't show you approved it

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Mac]

PR #746, opened 2026-01-21
fix(prs): Make r/R refresh properly update reviewer status as expected

sideshowbarker:
## Summary

Previously, `FetchPullRequest` and `FetchIssue` used a separate `cachedClient` with a 5-minute GraphQL cache TTL. That meant that after someone approved a PR, pressing “r” or “R” to refresh in gh-dash wouldn’t update the reviewer status — the enriched PR data was still being served from cache.

This PR fixes the problem by removing the `cachedClient` entirely and having `FetchPullRequest` and `FetchIssue` use the same shared `client` as the list-fetch functions (`FetchPullRequests`/`FetchIssues`). Each call now gets fresh data from GitHub.

Fixes https://github.com/dlvhdr/gh-dash/issues/745

`go test ./...` passes

### DIFF
--- a/internal/data/issueapi.go
+++ b/internal/data/issueapi.go
@@ -150,8 +150,8 @@ type IssuesResponse struct {
 // FetchIssue fetches a single issue by its GitHub URL
 func FetchIssue(issueUrl string) (IssueData, error) {
 	var err error
-	if cachedClient == nil {
-		cachedClient, err = gh.NewGraphQLClient(gh.ClientOptions{EnableCache: true, CacheTTL: 5 * time.Minute})
+	if client == nil {
+		client, err = gh.DefaultGraphQLClient()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


dlvhdr:
I honestly don't know what's our use case for caching a PR for 5 minutes. I can only think of issues it might cause. WTYT about removing it? am I missing something?

sideshowbarker:
> I honestly don't know what's our use case for caching a PR for 5 minutes. I can only think of issues it might cause. WTYT about removing it? am I missing something?

The GraphQLClient calls are relatively slow, and they count against the user’s GH rate-limit quota. But that only matters if we’re making relatively a lot of them, and often. Neither of which we’re doing — so yeah, I’ll make an update here that just removes it.

sideshowbarker:
Update pushed

PR #752, opened 2026-01-27
feat: add OSC52 clipboard support with DA1 detection

aredridel:
# Summary

Adds native terminal clipboard handling so it works remotely and in zellij.

Tested in Zellij and manually in Ghostty; tested non-OSC52 case in MacOS terminal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@
 - Custom actions to perform your specific workflow needs
 - Everything you can do on GitHub - diff, comment, checkout, push, update etc.
 - Control every setting with a YAML config file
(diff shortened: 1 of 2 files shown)


dlvhdr:
hey, thanks for the PR. The upcoming bubbletea v2 will support osc52, so I think we should wait on it. See https://github.com/charmbracelet/bubbletea/blob/v2-exp/clipboard.go

dlvhdr:
Closing due to lack of response, if you still think this is relevant, feel free to reopen

aredridel:
Oh! Sorry! thought I had — waiting seems reasonable enough given that!

Issue #753, opened 2026-01-27
[BUG] After viewing item in Notifications you can't switch to Issues or PRs (v4.22.0)


sdwheeler:
**Describe the bug**

**To Reproduce**
1. Hit <kbd>ss</kbd> to switch to Notifications view
2. Select an item and hit <kbd>Enter</kbd> to view it
3. Hit <kbd>s</kbd> to switch back to Issues or PRs

The view doesn't change.

**Expected behavior**
Should be able to switch views.

**Desktop (please complete the following information):**
- Windows 11
- dash 4.22.0
- GH CLI v2.86.0 (2026-01-21)

sideshowbarker:
I can reproduce this and I’ve got a fix written up for it. Will open a PR for it soon.

PR #754, opened 2026-01-27
fix: allow switching views with “s” key when viewing notifications

sideshowbarker:
## Summary
- Fix `s` key not working to switch views when viewing a PR or Issue notification in the Notifications dashboard

## Problem
When viewing a notification's details (PR or Issue) in the Notifications dashboard, pressing `s` to switch to the PRs or Issues view didn't work. The view would stay on Notifications.

## Cause
The PR and Issue notification handlers in the switch statement would run first when viewing a notification, and since Go's switch doesn't fall through by default, the fallback `s` key handler was never reached.

## Solution
Added explicit `s` key handlers in both the PR and Issue notification cases, to properly switch views. Also added `ClearSubject()`, to clear cached notification data when leaving the Notifications view.

Fixes https://github.com/dlvhdr/gh-dash/issues/753

### DIFF
--- a/internal/tui/ui.go
+++ b/internal/tui/ui.go
@@ -348,18 +348,7 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m, cmd
 
 			case key.Matches(msg, keys.BranchKeys.ViewPRs):
-				m.ctx.View = m.switchSelectedView()
-				m.syncMainContentWidth()
-				m.setCurrSectionId(m.getCurrentViewDefaultSection())
-
-				currSections := m.getCurrentViewSections()
-				if len(currSections) == 0 {
-					newSections, fetchSectionsCmds := m.fetchAllViewSections()
-					currSections = newSections
-					cmd = fetchSectionsCmds
-				}
-				m.setCurrentViewSections(currSections)
-				cmds = append(cmds, m.onViewedRowChanged())
+				cmds = append(cmds, m.switchSelectedView())
 			}
 		case m.ctx.View == config.PRsView:
 			switch {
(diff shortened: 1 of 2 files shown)


internal/tui/ui.go
```diff
@@ -542,10 +542,34 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 					}
 				}
 
-				// Always sync and return after updating prView - needed for tab navigation
-				// which updates carousel state but doesn't return a command
-				m.syncSidebar()
-				return m, prCmd
+				// Handle 's' key to switch views
+				if key.Matches(msg, keys.PRKeys.ViewIssues) {
```
dlvhdr, line 511: Just confirmed and it seems like this code bit and the existing ones are exact copies. So I think it's safe to just move all that code into the `m.switchSelectedView` func.

```go
					m.ctx.View = m.switchSelectedView()

          // All the below into the above :)
					m.syncMainContentWidth()
					m.setCurrSectionId(m.getCurrentViewDefaultSection())

					currSections := m.getCurrentViewSections()
					if len(currSections) == 0 {
						newSections, fetchSectionsCmds := m.fetchAllViewSections()
						currSections = newSections
						cmds = append(cmds, m.tabs.SetAllLoading()...)
						cmd = fetchSectionsCmds
					}
					m.setCurrentViewSections(currSections)
					cmds = append(cmds, m.onViewedRowChanged())
```

sideshowbarker: > @dlvhdr wrote:
>
> Just confirmed and it seems like this code bit and the existing ones are exact copies. So I think it's safe to just move all that code into the `m.switchSelectedView` func.

Yup — pushed an update.

PR #199, opened 2022-11-10
fix: #124 exclude pull requests of archived repositories

yokishava:
# Summary

I have seen this issue. I added a change to exclude PRs in archived repositories.

https://github.com/dlvhdr/gh-dash/issues/124

I would have liked to be able to remove PRs from archived repositories in the Github API arguments, but after checking the documentation I found it difficult to do so. I have added an implementation to remove pull requests from archived repositories after fetching pull requests.

## How did you test this change?

I wrote unit tests for the added functions.

I ran the command `gh dash` to see if it was successful in removing pull requests from archived repositories.

I used this archived repository.
https://github.com/yokishava/k8s-manifest-ci-demo

#### before
スクリーンショット 2022-11-11 1 47 38

#### after
スクリーンショット 2022-11-11 1 48 08

Issue #124, opened 2022-05-06
[BUG] dash should not lists PRs on archived repositories


obourgain:
**Describe the bug**
Dash lists the PRs on archived repos. I can't change those PRs, even not close them. I don't have the rights to unarchive the repos, so I just have zombies PRs in the list.

**To Reproduce**
1. Create a repo
2. Create a PR on it
3. Dash lists the repo
4. Archive the repo
5. Dash stills display the PR

**Expected behavior**
PRs on archived repos should be ignored.

**Desktop (please complete the following information):**
 - OS: OSX

dlvhdr:
Fixed by #124

PR #757, opened 2026-01-28
fix: exclude archived from TotalCount

### LINKED ISSUE #756: [BUG] Count mismatch query results when results include archived repositories
**Describe the bug**
Result list to a filter criteria does not include archived repositories. When I match the query in GitHub search the results include the archived repositories.

**To Reproduce**
1. specify a filter criteria where results include archived repositories
2. List does not include archived repositories.

**Expected behavior**
Result list and count match.

**Screenshots**

Image

Image

 - OS: macOS Tahoe 26.2

andrewg-xyz:
# Summary

  - Fix count mismatch when search results include archived repositories
  - Add archived:false to GitHub search queries for both PRs and issues

In #124 and #221, archived repositories were excluded from results because users can't interact with PRs/issues on archived repos. The fixes in #199 and #223 filtered archived repos client-side after fetching results, but the TotalCount returned from GitHub's API still included them, causing a mismatch between the displayed count and actual items.

- `task lint`
- `task test`
- build and run with custom gh-dash config and manually verify changes.

## Images/Videos
image
image

Fixes #756

### DIFF
--- a/docs/src/content/docs/configuration/issue-section.mdx
+++ b/docs/src/content/docs/configuration/issue-section.mdx
@@ -29,7 +29,9 @@ the issues view.
 This setting defines the [GitHub search filters][01] for the Issues in the section's table.
 
 Don't specify `is:issue` for this setting. The dashboard always adds that filter for Issues
-sections.
+sections. 
+
+Note issues from archived repositories are excluded by default with `archived:false`.
 
 You can define any combination of search filters. To make it easier to read and maintain
 your filters, we recommend using the `>-` syntax after the `filter` key and writing one
--- a/docs/src/content/docs/configuration/pr-section.mdx
+++ b/docs/src/content/docs/configuration/pr-section.mdx
@@ -31,6 +31,8 @@ This setting defines the [GitHub search filters][01] for the PRs in the section'
 Don't specify `is:pr` for this setting. The dashboard always adds that filter for PR
 sections.
 
+Note pull requests from archived repositories are excluded by default with `archived:false`.
+
 You can define any combination of search filters. To make it easier to read and maintain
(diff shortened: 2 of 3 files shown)


dlvhdr:
Hey, thanks for the PR :)
I'm wondering, is there ever a reason to search for prs/issues on archived repos?
Like, what if you're the one that archived the repo, maybe you still like to see prs/issues?
I need to read about what an archived repo actually means.
If we add the `is:archived` filter, users aren't aware, as the search bar shows just `is:pr`.
Now I'm not a fan of also showing `archived:false` as it takes up too much space but we can add a note in the docs site.

andrewg-xyz:
> Hey, thanks for the PR :) I'm wondering, is there ever a reason to search for prs/issues on archived repos? Like, what if you're the one that archived the repo, maybe you still like to see prs/issues? I need to read about what an archived repo actually means. If we add the `is:archived` filter, users aren't aware, as the search bar shows just `is:pr`. Now I'm not a fan of also showing `archived:false` as it takes up too much space but we can add a note in the docs site.

I had a similar debate, but made the assumption based on prior work. Archived repos are view-only. When I figured out why there was a difference, I couldn't even close the old PR. 

By default I would expect behavior to match that of GitHub's search. Considering the feature set of gh-dash, "view and take action" I like customizing the behavior to exclude archived repos since they aren't actionable. I could always view those items in GitHub UI if I really wanted to.

I'm happy to change behavior and/or update docs! But will wait for your opinion, I'm just a lowly user 😂 .

dlvhdr:
ok cool makes sense. Let's just add a disclaimer in the docs

dlvhdr (approved):

PR #759, opened 2026-02-01
Add mentions autocompletions

mchisolm0:
# Summary
- Add labeling and label completion to PRs
- Add autocomplete for mentioning users in comments
- Add autocompletion for users while assigning

- I built the extension and tested each of the features manually. 
- I ran `go test ./...` to run all the tests

### New Video
<video src="https://github.com/user-attachments/assets/adca505f-a9f9-45d0-86a0-1becc692276f">Unable to load video</video>

### Old Video
<video src="https://github.com/user-attachments/assets/306ef207-e653-4564-8fd7-9016ed66d6b4">Unable to load video</video>

### DIFF
--- a/internal/tui/components/prview/prview.go
+++ b/internal/tui/components/prview/prview.go
@@ -256,7 +256,10 @@ func (m *Model) renderLabels() string {
 		m.ctx.Styles.Common.MainTextStyle.Underline(true).Bold(true).Render(
 			fmt.Sprintf("%s Labels", constants.LabelsIcon)),
 		"",
-		common.RenderLabels(width, labels, style),
+		common.RenderLabels(labels, common.LabelOpts{
+			Width:     width,
+			PillStyle: style,
+		}),
 	)
 }
 
--- a/docs/src/content/docs/configuration/layout/pr.mdx
+++ b/docs/src/content/docs/configuration/layout/pr.mdx
@@ -27,6 +27,7 @@ By default, PR views display the following columns in the order they're listed:
 1. [`repo`] with a width of 15 columns.
 1. [`title`], set to grow to fill available space.
 1. [`author`] with a width of 10 columns.
+1. [`labels`] with a width of 22 columns, hidden by default.
 1. [`numComments`] with a width of 3 columns.
 1. [`reviewStatus`] with a width of 3 columns.
 1. [`ci`] with a width of 3 columns.
@@ -46,6 +47,7 @@ By default, PR views display the following columns in the order they're listed:
 [`repo`]:         #pr-repo-column
 [`title`]:        #pr-title-column
 [`author`]:       #pr-author-column
+[`labels`]:       #pr-labels-column
 [`numComments`]: #pr-number-of-comments-column
 [`reviewStatus`]: #pr-review-status-column
 [`ci`]:           #pr-continuous-integration-column
@@ -60,6 +62,9 @@ title:
   grow: true
 author:
   width: 15
+labels:
+  width: 22
+  hidden: true
 assignees:
   width: 20
   hidden: true
@@ -226,6 +231,17 @@ The heading for this column is <NerdFontIcon icon="nf-oct-checklist"/>
 [`theme.colors.text.faint`]: /configuration/theme#faint-text-color
 [`theme.colors.text.primary`]: /configuration/theme#primary-text-color
 [`theme.colors.text.warning`]: /configuration/theme#warning-text-color
+
+## PR Labels Column
+
+| Property | Type | Default                                                           |
+| :------- | :--- | :---------------------------------------------------------------- |
+| `labels` | yaml | <Code code={`width: 22\nhidden: true`} lang="yaml" frame="none"/> |
+
+This column displays the labels assigned to a PR as colored pills.
+Up to 6 labels per PR are fetched and displayed.
+
+The heading for this column is <NerdFontIcon icon="nf-md-label_multiple_outline"/>.
 
 # PR Lines Column
 
--- a/docs/src/data/schemas/layout/pr.yaml
+++ b/docs/src/data/schemas/layout/pr.yaml
@@ -28,6 +28,7 @@ schematize:
       1. [sref:`repo`] with a width of 15 columns.
       1. [sref:`title`], set to grow to fill available space.
       1. [sref:`author`] with a width of 10 columns.
+      1. [sref:`labels`] with a width of 22 columns, hidden by default.
       1. [sref:`reviewStatus`] with a width of 3 columns.
       1. [sref:`ci`] with a width of 3 columns.
       1. [sref:`lines`] with a width of 16 columns.
@@ -48,6 +49,7 @@ schematize:
       [sref:`repo`]:         layout.pr.repo
       [sref:`title`]:        layout.pr.title
       [sref:`author`]:       layout.pr.author
+      [sref:`labels`]:       layout.pr.labels
       [sref:`reviewStatus`]: layout.pr.reviewStatus
       [sref:`ci`]:           layout.pr.ci
       [sref:`lines`]:        layout.pr.lines
@@ -58,6 +60,9 @@ default:
     width: 15
   author:
     width: 15
+  labels:
+    width: 22
+    hidden: true
   assignees:
     width: 20
     hidden: true
@@ -183,6 +188,15 @@ properties:
       format: yaml
       details: |
         This setting affects display of the role icon for the person who created the PR.
+  labels:
+    title: PR Labels Column
+    description: Defines options for the labels column in a PR section.
+    type: object
+    oneOf:
+      - $ref: ./options.yaml
+    default:
+      width: 22
+      hidden: true
   assignees:
     title: PR Assignees Column
     description: Defines options for the assignees column in a PR section.
--- a/docs/src/pages/schema/layout/pr.json.ts
+++ b/docs/src/pages/schema/layout/pr.json.ts
@@ -16,6 +16,10 @@ export function GET() {
         author: {
           width: 15,
         },
+        labels: {
+          width: 22,
+          hidden: true,
+        },
         assignees: {
           width: 20,
           hidden: true,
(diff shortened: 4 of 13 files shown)


internal/data/userapi.go
```diff
@@ -0,0 +1,127 @@
...
+
+	log.Debug("FetchRepoUsers: executing GraphQL query", "repo", repoNameWithOwner)
+
+	// Query only publicly available mentionable users
+	// This includes anyone who has interacted with the repo (issues, PRs, comments)
+	var result MentionableUsersResponse
+	variables := map[string]any{
+		"owner": graphql.String(owner),
+		"name":  graphql.String(name),
+		"limit": graphql.Int(100),
```
dlvhdr, line 60: 100 will probably be fine for most repos, but maybe we can do better for the other repos too. It's up to you if you want to implement some sort of pagination here or in a future PR. GitHub themselves seem to return all nodes because they're priviliged and don't limit themselves on the BE like us plebs :D
If I look at this PR https://github.com/golang/go/pull/77492 and start typing `@` they fetch all the 2300 contributors

internal/data/userapi.go
```diff
@@ -0,0 +1,127 @@
...
+	graphql "github.com/cli/shurcooL-graphql"
+)
+
+var (
+	repoUserCache = make(map[string][]User)
+	userCacheMu   sync.RWMutex
+)
+
+type User struct {
+	Login string `json:"login"`
```
dlvhdr, line 16: I think we can add a nice QOL feature by also fetching the name of the user.
I have some coworkers with really hard to remember user names :D
So in the autocomletion we can show in a faint color they actual names.
Something like `@dlvhr Dolev Hadar`. We can even copy the style from github where the user name is bold.

internal/data/userapi.go
```diff
@@ -0,0 +1,127 @@
...
+// anyone who can interact with the repository (issue/PR authors, commenters, etc.)
+func FetchRepoUsers(repoNameWithOwner string) ([]User, error) {
+	// Check cache first
+	if cachedUsers, ok := CachedRepoUsers(repoNameWithOwner); ok {
+		log.Debug("FetchRepoUsers: cache hit", "repo", repoNameWithOwner, "count", len(cachedUsers))
+		return cachedUsers, nil
+	}
+	log.Debug("FetchRepoUsers: cache miss", "repo", repoNameWithOwner)
+
+	parts := splitRepoName(repoNameWithOwner)
```
dlvhdr: we can clean this up by also fetching the owner of the repo. We already do that with the name but not the owner for some reason. See https://github.com/dlvhdr/gh-dash/blob/main/internal/data/repository.go?plain=1#L17

internal/tui/components/autocomplete/utils.go
```diff
@@ -0,0 +1,373 @@
...
+	Word     string
+	StartIdx int
+	EndIdx   int
+	IsFirst  bool
+	IsLast   bool
+}
+
+// LabelContextExtractor extracts label context from a comma-separated list
+// at the given cursor position. Returns the current label text and its start/end positions.
+func LabelContextExtractor(input string, cursorPos int) (context string, start int, end int) {
```
dlvhdr: I'm thinking these shouldn't live in the `autocomplete` package. Instead, we can create an interface like:
```go
type Context struct {
  Start int
  End int
  Content string
}

interface Autocompleter {
  ExtractContext(input string, cursorPos int) Context
  InsertSuggestion(input string, suggestion string, contextStart int, contextEnd int) (newInput string, newCursorPos int)
  ItemsToExclude(input string, cursorPos int) []string
}
```

Having the autocomplete know about labels feels off. So it's just to make a bit more extendable and clean, nothing major..

internal/tui/components/prview/prview.go
```diff
@@ -26,11 +29,37 @@ import (
 var (
 	htmlCommentRegex = regexp.MustCompile("(?U)<!--(.|[[:space:]])*-->")
 	lineCleanupRegex = regexp.MustCompile(`((\n)+|^)([^\r\n]*\|[^\r\n]*(\n)?)+`)
-	commentPrompt    = "Leave a comment..."
-	approvalPrompt   = "Approve with comment..."
 	foldBodyHeight   = 8
 )
 
+// RepoLabelsFetchedMsg is sent when repository labels are successfully fetched
+type RepoLabelsFetchedMsg struct {
```
dlvhdr: maybe there's some reuse we can do between issueview and here as both contain a large chunk of lines added. Haven't gone deep into these changes but just pops out

internal/tui/keys/prKeys.go
```diff
@@ -50,6 +51,10 @@ var PRKeys = PRKeyMap{
 		key.WithKeys("A"),
 		key.WithHelp("A", "unassign"),
 	),
+	Label: key.NewBinding(
```
dlvhdr, line 55: thanks for adding it!

dlvhdr (commented):
damn man this is so awesome!

internal/data/userapi.go
mchisolm0: If you are good with it, I think I would prefer waiting for another PR. I could see pagination letting us "scroll" the results when you move down past the bottom visible result (rather than wrapping back to the top like we do now) so you can access more than only the top four results.

I would want to wait so we could add the same functionality to labels at the same time. Thoughts?

internal/data/userapi.go
mchisolm0: I think I've got this working pretty well. I have it structured so I should be able to add details to label suggestions in a future PR pretty easily, too.

dlvhdr: That's awesome :)

internal/data/userapi.go
dlvhdr: Yeah I'm fine with adding pagination later, though I just checked and seems like we can do a significant improvement that might make pagination redundant.

internal/data/userapi.go
mchisolm0: Okay...so I didn't get it pushed last night like I wanted...but I did go ahead and include label details when I addressed the other feedback.

internal/tui/components/autocomplete/utils.go
mchisolm0: I think I have this much cleaner now. Please give any feedback if there is anything else I need to improve. Thanks!

dlvhdr:
Looks really clean @mchisolm0!
I remember you wanted to push more changes, right? like moving the sources to a separate module?

Asking since I want to work on migrating to bubbletea v2 and we will have merge conflicts. I want to either push your changes first then fix the conflicts myself, or have you push your changes to your branch and then I can rebase on top of you.

mchisolm0:
> I remember you wanted to push more changes, right? like moving the sources to a separate module?

I might be misunderstanding, but I think I already made that change in https://github.com/dlvhdr/gh-dash/pull/759/commits/ac9a866d58656bc0f01311d063ee60b9f96bbde4 by moving the source implementations into  `internal/data/autocomplete`.

Is that the change you meant by “move sources to a separate module”?
- If yes, do you want the package renamed (for example, `internal/data/sources`)?
- If not, can you clarify what you meant?

Happy to make the adjustments. 😁

> Asking since I want to work on migrating to bubbletea v2

🎉

dlvhdr:
Ah ok my bad :) I'll try and merge this today then

mchisolm0:
No worries (: I will mark it ready for review. As a note for me to remember, I will want to make 2 new issues and then PRs for:
1. The pagination (since I haven't had time to dig into that just yet), and
2. The styling of the detail column of the suggestion popup (currently, they seem to both be the same color...I think because overriding the foreground color in the popup style)

Here is a video of 2

<details><summary>Screen recording of popup text being the same color</summary>

https://github.com/user-attachments/assets/fecfd712-4e1c-4b31-bab9-ec00b58f225c


<video src="">Unable to load video</video>
</details>

sideshowbarker:
https://github.com/sideshowbarker/gh-dash/commit/3b519c20db6f9109f39ba1f2dc30b037555e361b has a fix for the CI failure, in the `TestNotificationView_PRViewTabNavigation` test.

```diff
diff --git a/internal/tui/ui_test.go b/internal/tui/ui_test.go
index 2385c73..d8dd21f 100644
--- a/internal/tui/ui_test.go
+++ b/internal/tui/ui_test.go
@@ -322,10 +322,12 @@ func TestNotificationView_PRViewTabNavigation(t *testing.T) {
 	sidebarModel.UpdateProgramContext(ctx)
 
 	m := Model{
-		ctx:     ctx,
-		keys:    keys.Keys,
-		prView:  prview.NewModel(ctx),
-		sidebar: sidebarModel,
+		ctx:              ctx,
+		keys:             keys.Keys,
+		prView:           prview.NewModel(ctx),
+		sidebar:          sidebarModel,
+		issueSidebar:     issueview.NewModel(ctx),
+		notificationView: notificationview.NewModel(ctx),
 	}
 
 	// Set up a PR notification subject so GetSubjectPR() returns non-nil
```
On Discord, we’d talked about the `TestAutocompleteResetOnAssignMode` test — but when I test locally with the code state from https://github.com/mchisolm0/gh-dash/commit/8c8028ad06d156405b66d6bbd85ded4ca8952a55, the `TestAutocompleteResetOnAssignMode` test doesn’t fail. Nothing fails except that `TestNotificationView_PRViewTabNavigation` test.

mchisolm0:
> the `TestAutocompleteResetOnAssignMode` test doesn’t fail.

At some point in my initial work on this PR, I apparently deleted `TestAutocompleteResetOnAssignMode`. The thinking was that it was no longer valid since assign mode will now have user suggestions. The better approach would have been to ask before deleting a test, so I wanted to ask what the preference is, now.

Also, thanks for the test fix. I will get that and push the change. 😁

sideshowbarker:
> At some point in my initial work on this PR, I apparently deleted `TestAutocompleteResetOnAssignMode`. The thinking was that it was no longer valid since assign mode will now have user suggestions. The better approach would have been to ask before deleting a test, so I wanted to ask what the preference is, now.

As far as I can see the `TestEnterAssignModeResetsAutocompleteState` test you added is essentially providing the same test coverage as the `TestAutocompleteResetOnAssignMode` test had been providing. So I think it’s all good now.

PR #762, opened 2026-02-06
feat: dynamic preview pane width based on terminal size

trolleksii:
# Summary

The preview pane width is currently configured as a static number. This works perfectly fine when terminal window size remains unchanged. However in cases where terminal window changes(eg tmux vertical splits, switching from the laptop screen to an ultra-wide external monitor, etc) the layout gets messed up which may be quite frustrating.

I attempted to add the capability to set preview pane width as a fraction. Now defaults.preview.width accepts floats where:

0 to 1 → treated as a fraction (e.g., 0.4 = 40% of terminal width)
Above 1 → converted to int and works as before (backward compatible)

Note: This is my first contribution to the project, so let me know if there's anything that should be adjusted to better match the overall architecture or coding conventions.

I added a couple of tests to cover the change plus manual testing.

## Images/Videos
Example with with set to 0.45
Fullscreen:
image
With vertical split:
image

### DIFF
--- a/internal/tui/ui.go
+++ b/internal/tui/ui.go
@@ -983,7 +983,7 @@ func (m *Model) syncMainContentWidth() {
 		w := m.ctx.Config.Defaults.Preview.Width
 		if w > 0 && w < 1 {
 			w *= float64(m.ctx.ScreenWidth)
-		} 
+		}
 		m.ctx.DynamicPreviewWidth = min(int(w), m.ctx.ScreenWidth)
 		sideBarOffset = m.ctx.DynamicPreviewWidth
 	}


docs/src/pages/schema/defaults.json.ts
```diff
@@ -63,9 +63,9 @@ export function GET() {
             width: {
               title: "Preview Pane Width",
               description:
-                "Specifies the width of the preview pane in columns.",
-              type: "integer",
-              minimum: 1,
+                "Specifies the width of the preview pane. Numbers between 0 and 1 represent size relative to overall tewminal window size(eg 0.4 is 40%), numbers >=1 represent size in columns.",
```
dlvhdr:
```suggestion
                "Specifies the width of the preview pane. Numbers between 0 and 1 represent size relative to overall terminal window size (e.g 0.4 is 40%), numbers >=1 represent size in columns.",
```

internal/config/parser.go
```diff
@@ -107,7 +107,7 @@ type NotificationsSectionConfig struct {
 
 type PreviewConfig struct {
 	Open  bool
-	Width int
+	Width float64
```
dlvhdr: Should we also validate that 0 is not supplied? Somethink like:
```go
Width  float64  `yaml:"width" validate:"gt=0"`
```

dlvhdr (commented):
looks really good! thinking about it, should we change the default width to be
percenrage based? I'm not sure what's a better default

internal/config/parser.go
trolleksii: That's a very good point.

trolleksii:
> looks really good! thinking about it, should we change the default width to be percenrage based? I'm not sure what's a better default

I also think that something close to 50% could be a good starting point. I didn't want to impose my preferences but if we agree on this matter I can adjust the default.

dlvhdr:
@trolleksii can we make it 0.45 as the default?

trolleksii:
Done

trolleksii:
@dlvhdr can you please re-run the ci. I messed up test file when pulled main in and resolved conflicts in github ui 🤦

dlvhdr (approved):

PR #767, opened 2026-02-09
feat(prs): show workflows awaiting approval in Checks tab and add V keybinding to approve them

sideshowbarker:
## Summary

- Add support for displaying "workflows awaiting approval" in the PR Checks tab — these are workflows that require manual approval before running (e.g., first-time contributor PRs)
- Show pending check suites (QUEUED/PENDING/WAITING status) in a dedicated "Pending" section
- Display required status checks from branch protection rules that haven't reported yet
- Add `V` keybinding to approve all workflows awaiting approval for the selected PR, with confirmation prompt and clear error messaging for non-approvable workflows

image

## Test plan

- [ ] View a PR from a first-time contributor that has workflows awaiting approval
- [ ] Verify the overview box shows "X awaiting approval" with the shield icon
- [ ] Verify the checks bar includes an orange/warning segment for awaiting approval
- [ ] Verify the "All Checks" section shows an "Awaiting Approval" header with workflow names
- [ ] View a PR with pending/queued check suites and verify they appear under "Pending"
- [ ] Press `V` on a PR with workflows awaiting approval, confirm with `y`, verify workflows get approved
- [ ] Press `V` on a PR where workflows aren't approvable (non-fork), verify clear error message
- [ ] Verify `V` keybinding appears in help (`?`) under PR actions as "approve all workflows"
- [ ] Run `go test ./...` — all tests pass

### DIFF
--- a/internal/tui/constants/constants.go
+++ b/internal/tui/constants/constants.go
@@ -39,7 +39,7 @@ const (
 	SuccessIcon          = ""
 	TeamIcon             = ""
 	WaitingIcon          = ""
-	ActionRequiredIcon   = "󰒃" // nf-md-shield_alert (warning shield)
+	ActionRequiredIcon   = "" // nf-cod-warning (matches GitHub UI)
 
 	BehindIcon         = "󰇮"
 	BlockedIcon        = ""


internal/data/prapi.go
```diff
@@ -178,6 +197,7 @@ type LastCommitWithStatusChecks struct {
 					}
 				} `graphql:"contexts(last: 100)"`
 			}
+			CheckSuites CheckSuites `graphql:"checkSuites(last: 20)"`
```
dlvhdr, line 205: When will check suites appear here and not in `StatusCheckRollup`?
Worth documenting it. Is it when there hasn't been a run because we didn't approve one?

Also, maybe this can be done more efficiently. I see that you can write this query:
```graphql
query ActionRequired {
  # url(pr...) - the pr query
  checkSuites(last: 20) {
    checkRuns(filterBy: {conclusions: [ACTION_REQUIRED]}) {
      # ...
    }
  }
}
```

internal/tui/components/prssection/prssection.go
```diff
@@ -101,6 +101,8 @@ func (m *Model) Update(msg tea.Msg) (section.Section, tea.Cmd) {
 						cmd = tasks.MergePR(m.Ctx, sid, pr)
 					case "update":
 						cmd = tasks.UpdatePR(m.Ctx, sid, pr)
+					case "approveWorkflows":
```
dlvhdr, line 104: I think it's an existing bug but the default is to approve running the workflows (capital Y), but pressing Enter with an empty input results in a no-op

internal/tui/constants/constants.go
```diff
@@ -39,6 +39,7 @@ const (
 	SuccessIcon          = ""
 	TeamIcon             = ""
 	WaitingIcon          = ""
+	ActionRequiredIcon   = "󰒃" // nf-md-shield_alert (warning shield)
```
dlvhdr: can we use the same one as in GitHub? ``

internal/data/prapi.go
sideshowbarker: > @dlvhdr wrote:
>
> When will check suites appear here and not in `StatusCheckRollup`?
> Worth documenting it. Is it when there hasn't been a run because we didn't approve one?
>

https://github.com/sideshowbarker/gh-dash/commit/2fd71d8ddfaf9453da378a7683e0cf3781b0f8e0

- Check suites appear here but not in `StatusCheckRollup` when the workflow hasn’t been approved to run yet.

- GitHub creates the `CheckSuite` (the container) when the PR is pushed, but doesn’t create any `CheckRun` objects inside it until the workflow actually starts executing.

- `StatusCheckRollup.contexts` is a union of `CheckRun` and `StatusContext` nodes — _not_ `CheckSuite` nodes — so workflows awaiting approval never show up there.

> Also, maybe this can be done more efficiently. I see that you can write this query:
> ```graphql
> query ActionRequired {
>   # url(pr...) - the pr query
>   checkSuites(last: 20) {
>     checkRuns(filterBy: {conclusions: [ACTION_REQUIRED]}) {
>       # ...
>     }
>   }
> }
> ```

Unfortunately it seems that ain’t gonna won’t work here — because `ACTION_REQUIRED` is the _check suite’s_ conclusion, not the _check runs’_ conclusion.

When a suite is `ACTION_REQUIRED`, it has _zero_ check runs (the workflow hasn’t started) — so a `checkRuns(filterBy: {conclusions: [ACTION_REQUIRED]})` would return nothing.

And `CheckSuiteFilter` only supports `appId` and `checkName` — no conclusion/status filtering — so we can’t narrow the suite query at the GraphQL level either.

internal/tui/components/prssection/prssection.go
sideshowbarker: > @dlvhdr wrote:
>
> I think it's an existing bug but the default is to approve running the workflows (capital Y), but pressing Enter with an empty input results in a no-op

Went ahead and pushed a fix for it on this branch:

https://github.com/sideshowbarker/gh-dash/commit/6aad8d79b1493ca8c9c362bad5e30a11ffc5a6f4

But if you think it’d be better to split that commit out to a separate branch, we could do that?

internal/tui/constants/constants.go
sideshowbarker: > @dlvhdr wrote:
>
> can we use the same one as in GitHub? ``

👍 Made it so.

https://github.com/sideshowbarker/gh-dash/commit/7c10612df6404932a50b4ef920fb6bd466b66f91

internal/data/prapi.go
dlvhdr: Got it. Thanks for the explanation

PR #769, opened 2026-02-11
fix: smart filter inside repo

iperzic:
# Summary

When running `gh-dash` inside a git repo with smart filtering enabled, removing the `repo:` filter from the search bar and pressing Enter had no effect, the filter was silently re-added behind the scenes on every query. This happened because `IsFilteredByCurrentRemote` was never cleared when the user edited the search value.

This PR adds `SyncSmartFilterWithSearchValue()` which detects that the user removed the repo filter and clears the flag, so the search respects the user's edit.

> [!NOTE]  
> The issue is only reproducible when `filters` is removed from "Mine" PR section in `.gh-dash.yml`

Unit tests covering: smart filter active with repo filter present, smart filter toggled off, user manually removing the repo filter, user replacing with a different repo, and config already containing a repo filter.

## Images/Videos

| Before | After |
|--------|--------|
| CleanShot 2026-02-11 at 10 20 38@2x | CleanShot 2026-02-11 at 10 18 34@2x |

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -1,0 +1,18 @@
+name: Build Docs
+
+on:
+  pull_request:
+    paths:
+      - "docs/**"
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout your repository using git
+        uses: actions/checkout@v6
+      - name: Install, build, and upload your site output
+        uses: withastro/action@v5
+        with:
+          path: ./docs
--- a/.github/workflows/publish-docs.yaml
+++ b/.github/workflows/publish-docs.yaml
@@ -1,9 +1,6 @@
-name: Build Docs Docker image
+name: Build and Push Docs Docker image
 
 on:
-  pull_request:
-    paths:
-      - "docs/**"
   push:
     branches: ["main"]
     paths:


dlvhdr:
Hey @iperzic. I always disliked how smart filtering behaved.
I've pushed some more fixes on top of yours - let me know what you think.

This is how I want it to behave, when toggling smart filtering:
- if we're filtering by the current repo, remove that filter
- if we're filtering by another repo - do nothing
- if we're not filtering by any repo - add the current repo as filter
- auto add the current repo as filter on launch if it's missing and `smartFilteringAtLaunch` is set to true (which is the default).

Let me know if my fixes make sense and I would love for more tests are that.
I kept your fix working, just broken it into a function.

CC @sideshowbarker

sideshowbarker:
There’s certainly always been a huge amount of room for improvement in the smart-filtering feature. So it’s great to see it getting some attention and fixes/refinements.

https://github.com/sideshowbarker/gh-dash/commit/2112c161f5b243e4a3ef5aaa5586d5debdf591bf has some further fixes, plus tests.

I don’t have push/write perms to this repo, nor to PR branches here like this one — so that commit is at my own repo. But it could be fetched and cherry-picked into this branch.

iperzic:
@dlvhdr Thanks for the review, and the follow-up. I included @sideshowbarker commit in this PR as well. 

As far as I'm concerned, this looks good. Though, I must caveat that I'm not that experienced with golang, so take my review with a bucket of salt.

dlvhdr (approved):

dlvhdr:
@sideshowbarker seems like the tests fail due to not running inside a git repo or something along those lines. Should we mock the currentRepoFilter func?

sideshowbarker:
> @sideshowbarker seems like the tests fail due to not running inside a git repo or something along those lines.

To fix that, I see you made this change:

```diff
 func currentRepoFilter(t *testing.T) string {
 	t.Helper()
+	t.Setenv("GH_REPO", "https://github.com/dlvhdr/gh-dash")
 	repo, err := repository.Current()
 	if err != nil {
 		t.Fatal("failed to resolve current repository:", err)
 ```
 
So yeah I hadn’t considered how that CI might not (wouldn’t) have a remote configured.
 
But in hindsight: I can see that, because the test as currently written isn’t testing across multiple repos anyway, then for the current scope of the test, the test didn’t need `currentRepoFilter` to begin with. We could just do this:
 
 ```go
 const testRepoFilter = "repo:dlvhdr/gh-dash"
 ```
 …and then just have the test use that constant everywhere it’s calling `currentRepoFilter` now.

> Should we mock the `currentRepoFilter` func?

No, I think the fix you made is sufficient for now. The actual app code doesn’t call `currentRepoFilter` anywhere. It’s just a test-only thing.

I had imagined the test needed to handle testing across multiple repos, and so we’d need some way for the test to know what the current repo is. But… the test as currently written isn’t testing across multiple repos — and even if it were, we could anyway actually just be using strings.

So _when_ (if) we next have need to go back in and touch that `section_test.go` test, then we should refactor it to replace the `currentRepoFilter` function and calls.

But otherwise, I think the fix you made is fine for now, and we don’t need to fiddle with it further right now.

PR #779, opened 2026-02-16
fix(notifications): add esc key to return to notification prompt

joelazar:
Hi @dlvhdr,

First of all, thanks for adding the notifications feature to the project. Super nice addition 🙌 I was working on it as well on my fork for a bit, but I never got it in a "nice" state enough for firing off for a pull request.

When I tried it out today, I got into a bit of a confusing state though, when I "viewed" one of the notifications and later moved on to another one, I wasn't able to interact with them anymore, because the keybindings were messed up after that subview.

I added another keybinding for being able to move back to the none view mode now and added the following changes:

- Lets users exit PR/Issue notification detail mode and return to the default notification prompt
- Resets notification subject/help state when leaving detail mode (and when subject is cleared during sync)

Let me know what you think.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -85,7 +85,7 @@ tasks:
     cmds:
       - cd docs && pnpm i
 
-  docs-dev:
+  docs:
     cmds:
       - cd docs && pnpm dev
     desc: Start docs server
--- a/docs/src/components/DiffnavMiniLogo.astro
+++ b/docs/src/components/DiffnavMiniLogo.astro
@@ -22,7 +22,7 @@ const { class: className } = Astro.props;
   <pre
     aria-label="DIFFNAV ASCII Logo"
     class:list={[
-      "w-fit text-xl leading-none lg:text-2xl",
+      "w-fit text-xl leading-none lg:text-lg",
       className,
     ]}>
 ▜▔▚ ▌▐▔▔▐▔▔▐▚ ▌▐▔▌▐ ▞
--- a/docs/src/components/EnhanceMiniLogo.astro
+++ b/docs/src/components/EnhanceMiniLogo.astro
@@ -22,7 +22,7 @@ const { class: className } = Astro.props;
   <pre
     aria-label="ENHANCE ASCII Logo"
     class:list={[
-      "text-xl leading-none lg:text-2xl",
+      "text-xl leading-none lg:text-lg",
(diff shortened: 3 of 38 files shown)


sideshowbarker:
> When I tried it out today, I got into a bit of a confusing state though, when I "viewed" one of the notifications and later moved on to another one, I wasn't able to interact with them anymore, because the keybindings were messed up after that subview.

Yeah that was my bug, but I believe we subsequently fixed that in 6f20475, right?

> _Lets users exit PR/Issue notification detail mode and return to the default notification prompt_

Big +1 — that’s a great refinement. Seems kind of essential, actually. It rightly should have been part of the initial implementation.

> _Resets notification subject/help state when leaving detail mode (and when subject is cleared during sync)_

That’s a nice improvement too.

As a user of the Notifications dashboard, those are both changes I’d like to have.

joelazar:
> > When I tried it out today, I got into a bit of a confusing state though, when I "viewed" one of the notifications and later moved on to another one, I wasn't able to interact with them anymore, because the keybindings were messed up after that subview.
> 
> Yeah that was my bug, but I believe we subsequently fixed that in [6f20475](https://github.com/dlvhdr/gh-dash/commit/6f204751d8cb88d4119f42bcf3c3403998e13c97), right?
> 
> > _Lets users exit PR/Issue notification detail mode and return to the default notification prompt_
> 
> Big +1 — that’s a great refinement. Seems kind of essential, actually. It rightly should have been part of the initial implementation.
> 
> > _Resets notification subject/help state when leaving detail mode (and when subject is cleared during sync)_
> 
> That’s a nice improvement too.
> 
> As a user of the Notifications dashboard, those are both changes I’d like to have.

Yes, sorry, I can confirm that commit was actually fixing that issue which I mentioned there, I was originally testing on the released version.

Thank you for the feedback (and for the feature 😉)

dlvhdr:
o]]hkjkjjk f;

dlvhdr:
this is great! can we also add that key under https://github.com/dlvhdr/gh-dash/blob/main/internal/tui/ui.go?plain=1#L1174, saying something like `Enter (and esc to go back)`

joelazar:
Done ✅ Added an `Esc  go back` entry right below the `Enter` key in the help section now

dlvhdr:
Thank you!

PR #780, opened 2026-02-17
fix: re-enable mouse tracking after external process exits

sideshowbarker:
## Summary

Bubble Tea’s `RestoreTerminal()` restores the alt screen, bracketed paste, and focus reporting after a `tea.ExecProcess` finishes — but it doesn’t restore mouse mode. That means: After running any external command (diff, merge, custom commands), mouse-click events stop working — because the terminal is no longer sending mouse-escape sequences to the app.

### Root cause

When `tea.ExecProcess` runs, Bubble Tea calls `ReleaseTerminal()` —which (correctly) disables mouse tracking, so the child process gets a clean terminal. When the child exits, `RestoreTerminal()` restores everything — except mouse mode. That is apparently a [known Bubble Tea bug](https://github.com/charmbracelet/bubbletea/issues/1424).

### Fix

Return `tea.EnableMouseCellMotion` as a command when handling messages from exec process callbacks:

- `execProcessFinishedMsg` – custom commands
- `ExecFinishedMsg` (new) – diff viewer (`gh pr diff`)
- `TaskFinishedMsg` – merge, create, and other background tasks

Also fixes `DiffPR()`, which previously returned `nil` on success — meaning the UI never learned the diff process had finished. It now returns an `ExecFinishedMsg`, so the handler can re-enable mouse tracking and refresh data.

- Unit tests verifying that `tea.EnableMouseCellMotion` is returned in the command batch for all three message types
- Manual testing: open a PR diff with `d`, exit, verify mouse clicks still work

### DIFF
--- a/internal/tui/common/diff.go
+++ b/internal/tui/common/diff.go
@@ -26,6 +26,6 @@ func DiffPR(prNumber int, repoName string, env []string) tea.Cmd {
 		if err != nil {
 			return constants.ErrMsg{Err: err}
 		}
-		return nil
+		return constants.ExecFinishedMsg{}
 	})
 }
--- a/internal/tui/constants/errMsg.go
+++ b/internal/tui/constants/errMsg.go
@@ -5,3 +5,6 @@ type ErrMsg struct {
 }
 
 func (e ErrMsg) Error() string { return e.Err.Error() }
+
+// ExecFinishedMsg is sent when an external process (tea.ExecProcess) finishes.
+type ExecFinishedMsg struct{}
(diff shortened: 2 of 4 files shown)


dlvhdr:
I need to try and upgrade to bubbletea v2 and see if this issue persists

sideshowbarker:
Clicking the **`donate`** button seems to not be working at all now (after the v2 upgrade) – so, I reckon I’ll turn this PR into a _“make mouse tracking work again”_ PR…

dlvhdr:
thanks on the previous PR @sideshowbarker, so now we need to rebase this one to bubbletea v2

sideshowbarker:
> thanks on the previous PR @sideshowbarker, so now we need to rebase this one to bubbletea v2

With https://github.com/dlvhdr/gh-dash/pull/804 merged, we don’t actually need this any longer — because it’s not reproducible with v2. So let’s go ahead just close this.

PR #782, opened 2026-02-17
fix: support custom keybindings for PR/Issue notifications

sideshowbarker:
**Problem:** Custom keybindings defined under `keybindings.prs` or `keybindings.issues` in any user config unexpectedly do not work when viewing PR/Issue notifications in the Notifications dashboard.

**Root cause:** Both `isUserDefinedKeybinding()` and `executeKeybinding()` only handle `PRsView`, `IssuesView`, and `RepoView` — the `NotificationsView` case is missing entirely.

**Fix:** This PR adds `NotificationsView` handling that checks the notification’s subject type, and dispatches to the appropriate keybinding set.

### Template variables

Template variables `RepoName` and `PrNumber`/`IssueNumber` are always available. When the sidebar is open (i.e., the user has pressed Enter to view the notification), additional fields are populated from the fetched PR/Issue data:

| State | PR template fields |
|-------|-------------------|
| Sidebar not open | `RepoName`, `PrNumber`, `RepoPath` |
| Sidebar open | + `HeadRefName`, `BaseRefName`, `Author` |

| State | Issue template fields |
|-------|---------------------|
| Sidebar not open | `RepoName`, `IssueNumber`, `RepoPath` |
| Sidebar open | + `Author` |

If a command template references a sidebar-only field before the sidebar is opened, the template engine’s `missingkey=error` option produces a clear error message.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -85,7 +85,7 @@ tasks:
     cmds:
       - cd docs && pnpm i
 
-  docs-dev:
+  docs:
     cmds:
       - cd docs && pnpm dev
     desc: Start docs server
--- a/docs/src/components/DiffnavMiniLogo.astro
+++ b/docs/src/components/DiffnavMiniLogo.astro
@@ -22,7 +22,7 @@ const { class: className } = Astro.props;
   <pre
     aria-label="DIFFNAV ASCII Logo"
     class:list={[
-      "w-fit text-xl leading-none lg:text-2xl",
+      "w-fit text-xl leading-none lg:text-lg",
       className,
     ]}>
 ▜▔▚ ▌▐▔▔▐▔▔▐▚ ▌▐▔▌▐ ▞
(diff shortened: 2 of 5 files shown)


dlvhdr:
Yesss thank you!

dlvhdr:
I think we should also support overriding notification view specific keybinds, like - mark as done etc, as part of this pr.
This is done in https://github.com/dlvhdr/gh-dash/blob/main/internal/tui/keys/keys.go?plain=1#L224. WDYT?

sideshowbarker:
> I think we should also support overriding notification view specific keybinds, like - mark as done etc, as part of this pr.
> This is done in https://github.com/dlvhdr/gh-dash/blob/main/internal/tui/keys/keys.go?plain=1#L224. WDYT?

👍 Yes — made it so: https://github.com/sideshowbarker/gh-dash/commit/51e8a8a0defbaad504b6aa62a24d38b9a81300cc

sideshowbarker:
> I think we should also support overriding notification view specific keybinds, like - mark as done etc, as part of this pr.
> This is done in https://github.com/dlvhdr/gh-dash/blob/main/internal/tui/keys/keys.go?plain=1#L224. WDYT?

👍 Yes — made it so: https://github.com/sideshowbarker/gh-dash/commit/51e8a8a0defbaad504b6aa62a24d38b9a81300cc

dlvhdr (approved):

PR #783, opened 2026-02-17
fix: notifications dashboard fails to display some notifications it should

sideshowbarker:
## Summary

**Bug:** The Notifications dashboard had been failing to show particular new notifications that it should be showing: Namely, new notifications in threads which include an older notification that a user marked Done.

**Cause:** GitHub notification thread IDs are per-subject (per issue/PR), not per-event. When a user marks a notification Done in gh-dash, the thread ID is stored in `done.json`. If new activity later arrives on that same thread, GitHub creates a new unread notification with the **same thread ID**. And because we had `IsDone()` only checking whether the ID existed in the store, **some new notifications will unexpectedly never get shown** — because they have a thread ID we had already marked as Done.

**Fix:** This PR replaces the current ID-only Done store with a **timestamp-aware Done store** that records *when* each notification was marked Done (using its `updated_at`). On subsequent fetches, `IsDone` compares the stored timestamp against the notification's current `updated_at`. If the notification has been updated since it was marked Done, it resurfaces.

Additionally, the Done store now **prunes stale entries on load**: entries older than 14 days and legacy zero-time entries are removed, keeping the file from growing indefinitely. GitHub only retains Done-marked notifications for about 1 week, so entries beyond 14 days are dead weight.

## ⚠️ Important: Breaking change for existing users

On first run after a dash upgrade with this change included, **dash will re-show users all of the notifications they have previously marked as Done**. The old format (`done.json` as a plain JSON array of IDs) stored no timestamps — so there's unfortunately **no way to determine which stored notification-thread IDs have new activity vs. which are genuinely stale.**

But that resurfacing of all Done notifications is a **one-time thing**. Now, once a user marks any notification as Done, a proper timestamp is stored for it — and then only any genuinely-new notification in its thread will ever be shown.

## How it works

| Scenario | Old behavior | New behavior |
|----------|--------------|--------------|
| Mark notification Done | ❌ Store only thread ID | ✅ **Change:** Store thread ID + `updated_at` timestamp |
| No new notification w/ same thread ID | ✅ Dash never again shows Done-marked notification | ✅ Same: Dash never again shows Done-marked notification |
| New notification w/ same thread ID | ❌ Bug: Dash does not show new notification | ✅ **Fixed:** Dash shows new notification |
| Legacy `done.json` on upgrade | N/A | ⚠️ **One time:** Dash (re)shows all existing Done-marked notifications from the Done store — as if they were new — and user needs to re-mark all those as Done ☹️ |
| Stale entries (>14 days old) | ❌ Accumulate forever | ✅ **New:** Pruned on startup |
| Legacy zero-time entries | ❌ Persist forever | ✅ **New:** Pruned on startup (functionally equivalent — `IsDone` returns false either way) |

### DIFF
--- a/internal/tui/components/notificationssection/DESIGN.md
+++ b/internal/tui/components/notificationssection/DESIGN.md
@@ -12,7 +12,9 @@ The notifications feature adds a new view to gh-dash that displays GitHub notifi
 internal/
 ├── data/
 │   ├── notificationapi.go       # GitHub API interactions for notifications
-│   └── bookmarks.go             # Local bookmark storage (singleton)
+│   ├── bookmarks.go             # Local bookmark storage (singleton)
+│   ├── donestore.go             # Timestamp-based Done tracking (singleton)
+│   └── donestore_test.go        # Tests for Done store
 ├── tui/
 │   ├── keys/
 │   │   └── notificationKeys.go  # Key bindings specific to notifications
@@ -148,7 +150,9 @@ Title truncation is handled dynamically by the table component based on actual c
 
 #### 8. Bookmark and Done Systems
 
-Both bookmarks and "done" status are tracked locally because GitHub's API doesn't provide these features. They share a common `NotificationIDStore` implementation in `data/bookmarks.go`:
+Both bookmarks and Done status are tracked locally because GitHub’s API doesn’t provide these features.
+
+**Bookmarks** use the generic `NotificationIDStore` type in `data/bookmarks.go`:
 
 ```go
 type NotificationIDStore struct {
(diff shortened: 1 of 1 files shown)


internal/data/donestore.go
```diff
@@ -0,0 +1,181 @@
...
+	if s.filePath == "" {
+		return nil
+	}
+
+	tsMap := make(map[string]string, len(s.entries))
+	for id, t := range s.entries {
+		tsMap[id] = t.Format(time.RFC3339)
+	}
+
+	data, err := json.MarshalIndent(tsMap, "", "  ")
```
dlvhdr: do we need an indent here? I don't think it should be readable.. also, maybe we can just save it in a more compact format - like `{id}<space>{timestamp}\n`. I'm a bit worried that over time this file would balloon.
Another question is, how long do we need to save these for?

sideshowbarker: > @dlvhdr wrote:
>
> I'm a bit worried that over time this file would balloon.
> Another question is, how long do we need to save these for?

You’re right. So I went ahead did https://github.com/sideshowbarker/gh-dash/commit/d9f687a5e8. From checking my own notifications, and asking a few other people, it seems that GH now only retains Done-marked notifications that are no more than one week old. So it seems like we can safely trim out any in our own Done store which are older than two weeks (one week + one extra week just for defensive safety).

I’d estimate than in most case for most normal users, that will keep the size of the Done store below about 200 entries or so. I’d be surprised if it ever grew to more than 2000 entries for anybody. I think somebody would really need to be getting an unmanageably huge number of notifications to begin with — for it to every grow that big.

But regardless, with this change, it’s no longer going to continue to grow unbounded. It’ll coalesce into some discrete bounded sized: Probably around 80K at the very most, and I would guess probably way less than 10K for most people.

> do we need an indent here? I don't think it should be readable

Right. I’d added that just to make it easier for me to inspect it while I was working on this. But yeah nobody but we would ever need to do that. Nor will I, even, after this lands.

So, switched it to just using `json.Marshal` (no indent).

> also, maybe we can just save it in a more compact format - like `{id}<space>{timestamp}\n`.

Your call. That’d be cleaner to read/inspect manually — but as mentioned above, we don’t really need that. And especially now we’re gonna be trimming the store, how much would that format buy in practice? It’d be the difference between this:

```
{"12345678":"2026-02-28T10:30:00Z","87654321":"2026-02-27T14:15:00Z"}
```

… and this:

```
12345678 2026-02-28T10:30:00Z
87654321 2026-02-27T14:15:00Z
```

In other words, the JSON format adds four `"` quote character per entry. Which yeah, is more than zero — but if a user has 100 items marked Done, the non-JSON was would still be several thousand characters: 3000, and then JSON way would be 3400. Either way, it’s still several thousand characters.

And the other tradeoff would be that we’d need to write and maintain code to manually parse the custom format. Which yeah admittedly would be very simple — though I think both the serialization and parsing sides would probably also need some custom integrity checking (to make sure it doesn’t end up writing out only part of line or whatever to the file). I mean, because we’d otherwise end up with a corrupted file — or risk reading in from broken data.

And that’s compared to just needing to call `json.Marshal` and `json.Unmarshal` that’ll handle everything for us.

dlvhdr: I'm totally fine with keeping using JSON.

> From checking my own notifications, and asking a few other people, it seems that GH now only retains Done-marked notifications that are no more than one week old. So it seems like we can safely trim out any in our own Done store which are older than two weeks (one week + one extra week just for defensive safety).

From what I understand, "done" notifications aren't shown again. We then trim the file and they aren't considered "done" anymore. Wouldn't they resurface? How does this work?

sideshowbarker: > From what I understand, "done" notifications aren't shown again. We then trim the file and they aren't considered "done" anymore. Wouldn't they resurface? How does this work?

They won’t resurface — because the notifications we trim will always only be two-week-old ones that no longer exist at GH. So we can’t re-retrieve they from GH. They’re gone.

When you mark a notification as Done, GH only retains it for a week. The API call is literally DELETE — and the GH docs provide no guarantees they will keep those DELETE-marked notifications around for any time _at all_.

So in practice _“Done”_ sort of means “Deleted” — if it weren’t for the fact the API provides no way to tell which ones have been DELETEd, and for the fact that they don’t actually delete them but instead keep them around for a while.

So I would assume the only reason GH keeps those DELETEd notifications around for any time at all is so that you can “undo” the deletion of any that you’ve _accidentally_ deleted. But it seems that in practice they only retain those for one week.

And so the code in this branch trims any notifications in the Done store that are older than two weeks. I _think_ we could probably trim them after a week, but I guess a bit of extra defensive safety makes sense. Maybe an entire extra week is excessive, I dunno. Maybe just making it 8 or 9 days is enough wiggle room.

But anyway, trimming those 2-week-old DELETEd notifications won’t ever cause any of them to resurface in Dash – since, as outlined above, the corresponding GH notifications would be long gone from GH by that time. So we won’t end up re-retrieving them.

dlvhdr: Got it. Thanks for the explanation!

PR #786, opened 2026-02-27
feat: add option to copy using osc52

### LINKED ISSUE #559: Feature: Tmux paste buffer and/or OSC 52 copies
I often use gh-dash on headless, remote machines. copy functionality, thus doesn't work since I'm not running X server for instance.

I use tmux paste buffer extensively. It would be neat if gh-dash could `tmux set buffer 'foo'` instead.

Falling back to OSC 52 would be neat. Or doing that, too. I let tmux set my laptop's clipboard via the same (enabled by my terminal emulator).

asla9709:
# Summary
  - Add \`oscClipboard\` config option to enable OSC 52 clipboard support for headless/remote machines
  - When enabled, the copy number (y) and copy url (Y) keybindings use OSC 52 escape sequences instead of system clipboard
  - Supports tmux and screen terminal multiplexers

  Fixes #559

  - [x] Enabled \`oscClipboard: true\` in config
  - [x] Verified copying works via tmux/screen to local clipboard
  - [x] Verified default behavior still uses system clipboard

# Images/Screenshots
N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -68,7 +68,7 @@ require (
 require (
 	github.com/alecthomas/chroma/v2 v2.19.0 // indirect
 	github.com/aymanbagabas/git-module v1.8.4-0.20231101154130-8d27204ac6d2
-	github.com/aymanbagabas/go-osc52/v2 v2.0.1 // indirect
+	github.com/aymanbagabas/go-osc52/v2 v2.0.1
 	github.com/aymerick/douceur v0.2.0 // indirect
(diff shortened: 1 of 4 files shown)


asla9709:
AI Disclaimer:
I wrote this manually a year ago, had Claude rebase it on the current repo.

dlvhdr:
I'm not really familiar with osc52 @asla9709. Any reason not to always use it?

PR #789, opened 2026-03-02
fix date syntax in docs

daturkel:
# Summary

The docs have several examples in the format `updated>=SOMEDATE`, but it should actually be `updated:>=SOMEDATE`.


dlvhdr:
Whoops! Thank you!

PR #838, opened 2026-04-15
fix: PrevItem does not scroll viewport at top boundary

lawrence3699:
**Bug:** When navigating up through a list that extends beyond one page, the viewport fails to scroll when the cursor reaches the topmost visible row. The cursor moves above the visible area, becoming invisible.

**Cause:** `PrevItem` uses `<` to check if the cursor is at the top of the viewport, but the symmetric `NextItem` correctly uses `>=` for the bottom boundary. Since `currId` is always `>= topBoundId` during normal step-by-step navigation, the `<` check never triggers when the cursor is at the top boundary.

**Before:** Cursor at row 1, `topBoundId=1`. Press up → cursor moves to row 0, viewport stays at `topBoundId=1`. Row 0 is above the visible area.

**After:** Same scenario → viewport scrolls up to `topBoundId=0`, cursor at row 0 is visible.

**Fix:** Change the check to `> 0 && <=` — the `<=` matches `NextItem`'s `>=` pattern, and the `> 0` guard prevents an unnecessary scroll when already at the first item.

**Validation:** Added regression tests for `PrevItem`/`NextItem` boundary scrolling and edge cases. `go test ./...` and `go build ./...` both pass.

**AI Disclosure:** Claude Code was used to identify the asymmetry and draft the fix and tests. I (the submitter) fully understand the off-by-one: `NextItem` scrolls when `currId >= bottomBoundId` (at or past the bottom row), but `PrevItem` only scrolled when `currId < topBoundId` (already past the top row, which doesn't happen in single-step navigation). The `<=` makes both directions scroll when the cursor is *at* the boundary.

### DIFF
--- a/internal/tui/components/listviewport/listviewport.go
+++ b/internal/tui/components/listviewport/listviewport.go
@@ -102,7 +102,7 @@ func (m *Model) NextItem() int {
 }
 
 func (m *Model) PrevItem() int {
-	atTopOfViewport := m.currId < m.topBoundId
+	atTopOfViewport := m.currId > 0 && m.currId <= m.topBoundId
 	if atTopOfViewport {
 		m.topBoundId -= 1
 		m.bottomBoundId -= 1
--- a/internal/tui/components/listviewport/listviewport_test.go
+++ b/internal/tui/components/listviewport/listviewport_test.go
@@ -1,0 +1,119 @@
+package listviewport
+
+import (
+	"testing"
+	"time"
+
+	"github.com/dlvhdr/gh-dash/v4/internal/tui/constants"
+	"github.com/dlvhdr/gh-dash/v4/internal/tui/context"
+)
+
+func newTestModel(numItems, viewportHeight, itemHeight int) Model {
+	ctx := context.ProgramContext{}
+	dims := constants.Dimensions{Width: 80, Height: viewportHeight}
+	now := time.Now()
+	return NewModel(ctx, dims, now, now, "items", numItems, itemHeight)
+}
+
+func TestPrevItemScrollsAtTopBound(t *testing.T) {
+	// 10 items, viewport height 5, item height 1 => 5 items per page
+	// Initial bounds: topBoundId=0, bottomBoundId=4
+	m := newTestModel(10, 5, 1)
+
+	// Navigate down past the first page boundary
+	// After 5 NextItem calls: currId=5, topBoundId=1, bottomBoundId=5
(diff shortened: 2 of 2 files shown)


internal/tui/components/listviewport/listviewport.go
```diff
@@ -102,7 +102,7 @@ func (m *Model) NextItem() int {
 }
 
 func (m *Model) PrevItem() int {
-	atTopOfViewport := m.currId < m.topBoundId
+	atTopOfViewport := m.currId > 0 && m.currId <= m.topBoundId
```
dlvhdr, line 105: I don't really get the need for the `m.currId > 0` condition. I tested it without it and it seems to work fine. Also the `atTopOfViewport` is misleading with that condition.

internal/tui/components/listviewport/listviewport_test.go
```diff
@@ -0,0 +1,119 @@
...
+	ctx := context.ProgramContext{}
+	dims := constants.Dimensions{Width: 80, Height: viewportHeight}
+	now := time.Now()
+	return NewModel(ctx, dims, now, now, "items", numItems, itemHeight)
+}
+
+func TestPrevItemScrollsAtTopBound(t *testing.T) {
+	// 10 items, viewport height 5, item height 1 => 5 items per page
+	// Initial bounds: topBoundId=0, bottomBoundId=4
+	m := newTestModel(10, 5, 1)
```
dlvhdr, line 21: can we add a struct for the arguements? makes it clearer 
```go
newTestModel(opts testModelOpts) Model {
  // ...
}
```

internal/tui/components/listviewport/listviewport_test.go
```diff
@@ -0,0 +1,119 @@
...
+	if m.GetCurrItem() != 5 {
+		t.Fatalf("expected currId=5 after 5 NextItem calls, got %d", m.GetCurrItem())
+	}
+	if m.topBoundId != 1 {
+		t.Fatalf("expected topBoundId=1, got %d", m.topBoundId)
+	}
+
+	// Navigate back up to the top bound
+	// 4 PrevItem calls: currId goes 4, 3, 2, 1
+	for i := 0; i < 4; i++ {
```
dlvhdr, line 38: maybe the `4` can be reused as `opts.viewportHeight - 1`?

dlvhdr (commented):
thanks for this! has been bugging me.
and also thx for adding tests!

Issue #812, opened 2026-03-18
[BUG] crashed on navigating multiple windows 


hawkaii:
**Describe the bug**
crashed  on navigating to issues by tapping `s` to switch

[debug.log](https://github.com/user-attachments/files/26079753/debug.log)

**To Reproduce**

Tap `s` to navigate multiple time on starting

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

- OS: Windows 11 (wsl: Ubuntu)
- Terminal Emulator: Wezterm
- Using tmux? Yes

sideshowbarker:
Can you look as see if there’s `bubbletea-panic-*.log` in your `/home/hawkaii/code/work/cabswale/raahi-new/raahi-app-controller` directory, and upload that here?

sideshowbarker:
And maybe also try `gh dash --debug 2>panic.log`. There should be some stack trace getting emitted to stderr that you can either see in your terminal or else redirect to a file like that.

hawkaii:
result of panic.log 
```
goroutine 1 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.25.7/x64/src/runtime/debug/stack.go:26 +0x64
runtime/debug.PrintStack()
	/opt/hostedtoolcache/go/1.25.7/x64/src/runtime/debug/stack.go:18 +0x1c
github.com/charmbracelet/bubbletea.(*Program).recoverFromPanic(0x40004ea000, {0x7c5b00, 0x12f1520})
	/home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.3.5/tea.go:810 +0x90
github.com/charmbracelet/bubbletea.(*Program).Run.func2()
	/home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.3.5/tea.go:601 +0xc4
panic({0x7c5b00?, 0x12f1520?})
	/opt/hostedtoolcache/go/1.25.7/x64/src/runtime/panic.go:783 +0x120
github.com/dlvhdr/gh-dash/v4/internal/tui.Model.Update({0x13299a0, {0x1, {0x0, 0x0}, {0x3b, 0x23, {{...}, {...}, {...}, {...}, ...}, ...}, ...}, ...}, ...)
	/home/runner/go/pkg/mod/github.com/dlvhdr/gh-dash/v4@v4.23.2/internal/tui/ui.go:239 +0x1504
[...]

PR #851, opened 2026-04-28
fix: guard isUserDefinedKeybinding against nil model context

### LINKED ISSUE #850: [BUG] panic: nil pointer dereference in isUserDefinedKeybinding when terminal sends OSC/CSI responses before model is initialized
# [BUG] panic: nil pointer dereference in `isUserDefinedKeybinding` when terminal sends OSC/CSI responses before model is initialized

**Describe the bug**

`gh dash` panics on launch with `runtime error: invalid memory address or nil pointer dereference` inside `(*Model).isUserDefinedKeybinding`. It happens when the host terminal proactively responds to bubbletea's startup queries (e.g. OSC 11 background-color, CSI cursor-position report) before the bubbletea program has finished setting up the model context. Those response bytes are parsed as `tea.KeyMsg` and dispatched to `Update`, which calls `isUserDefinedKeybinding(msg)` while `m.ctx` is still nil.

**Reproduction**

I see this consistently when launching `gh dash` from a [Wave Terminal](https://github.com/wavetermdev/waveterm) widget block (Wave eagerly emits color/cursor responses on tty open). The same panic has been reported elsewhere from xterm-style terminals that respond fast to the startup queries.

```
gh dash version dev
goos: darwin
goarch: arm64
module version: v4.22.0
module version: v4.23.2  (also reproduces)
```

**Stack trace (v4.22.0)**

```
panic: runtime error: invalid memory address or nil pointer dereference

github.com/dlvhdr/gh-dash/v4/internal/tui.(*Model).isUserDefinedKeybinding(...)
	internal/tui/ui.go:1416 +0x44
github.com/dlvhdr/gh-dash/v4/internal/tui.Model.Update(...)
	internal/tui/ui.go:208 +0xc1c
github.com/charmbracelet/bubbletea.(*Program).eventLoop(...)
	bubbletea@v1.3.5/tea.go:525
```

(v4.23.2 stack is identical with line numbers shifted: `ui.go:1459` and `ui.go:219`. Visible bytes preceding the panic include `^[]11;rgb:0000/0000/0000^[\` and `^[[1;1R`, confirming the rogue input is the OSC 11 / cursor-pos response.)

**Cause**

`isUserDefinedKeybinding` dereferences `m.ctx.Config.Keybindings.*` without checking whether `m.ctx` is nil. The `Update` loop can be entered with key messages before the context is wired up, when the terminal injects synthetic-looking input early.

**Suggested fix**

Guard the entry of `isUserDefinedKeybinding` with a nil-check on `m.ctx` and `m.ctx.Config`. PR with the 3-line guard incoming.

**Notes on AI usage** (per AI_POLICY.md)

I diagnosed and patched this with help from Claude Code. The fix is a single nil-guard at the top of the function — I have read it, can defend it, and built and verified the patched binary locally before opening this report.

qveys:
## Summary

Fixes #850 — guards `isUserDefinedKeybinding` against a nil `m.ctx` / `m.ctx.Config`, so terminals that emit OSC/CSI responses (e.g. Wave Terminal's eager OSC 11 background-color and CSI cursor-position reports) before the bubbletea model is fully initialized no longer panic the TUI.

## Cause

`Update` can dispatch `KeyMsg` to `isUserDefinedKeybinding` before `m.ctx` is wired up. The function then derefs `m.ctx.Config.Keybindings.Universal` and panics. This is the same class of issue as #812 / `f0b67d7`, but in a different code path that the previous fix didn't cover.

## Fix

Three-line guard at the top of `isUserDefinedKeybinding`:

```go
if m.ctx == nil || m.ctx.Config == nil {
    return false
}
```

Returning `false` here is the safe default: no user-defined binding can match before the config is loaded, so the regular built-in key dispatch handles the message normally (and at this point the relevant code paths have their own `currSection != nil` guards from `f0b67d7`).

## Test plan

- [x] Reproduced the panic on `main` and on tagged v4.22.0 / v4.23.2 from a Wave Terminal block (panic on first key event).
- [x] Built the patched binary, replaced the gh-extension binary, panic gone, `gh dash` opens normally and keybindings work.
- [ ] Maintainers: please confirm the fix doesn't regress user-defined keybinding lookups in normal startup paths.

## AI disclosure (per AI_POLICY.md)

Diagnosis and patch authored with Claude Code assistance. I (the submitter) read the function, understand why the nil guard is correct, and have built and tested the binary locally. Happy to iterate on review feedback.

### DIFF
--- a/internal/tui/ui.go
+++ b/internal/tui/ui.go
@@ -1630,6 +1630,9 @@ func (m *Model) switchSelectedView() tea.Cmd {
 }
 
 func (m *Model) isUserDefinedKeybinding(msg tea.KeyMsg) bool {
+	if m.ctx == nil || m.ctx.Config == nil {
+		return false
+	}
 	for _, keybinding := range m.ctx.Config.Keybindings.Universal {
 		if keybinding.Builtin == "" && keybinding.Key == msg.String() {
 			return true


Shemse1:
if m.ctx == nil || m.ctx.Config == nil {
    return false
}

Shemse1 (commented):
done

dlvhdr (approved):

PR #860, opened 2026-05-04
fix: use cmd.exe fallback for custom commands on Windows (closes #686)

### LINKED ISSUE #686: custom commands dont work in windows
**Describe the bug**

if i put the following in my config file :
```
keybindings:
  universal: []
  issues: []
  prs:
    - key: k
      name: squash
      command: gh pr --repo {{.RepoName}} {{.PrNumber}} -s
```

when i press `k` on the approved PR, i get `Whoops, got an error: exec: "sh": executable file not found in %PATH%`

this is because on windows `sh` nor `SHELL` ENV do not exist by default. 
see code `executeCustomCommand`

**To Reproduce**

1. add above to config
2. press k, on windows

**Expected behavior**
i expect the PR to be squashed and merged when k is pressed.
i expect custom commands on windows to work

 - OS: windows 11 in powershell

my main request is i want a way to squash and merge, and not just merge a PR.
but, besides that, it seems like the above keybinding should be made to work.
on windows, i think the logic should be to default to powershell

Booyaka101:
### Summary

Custom keybinding commands fail on Windows with `exec: "sh": executable file not found in %PATH%` because `executeCustomCommand` falls back to `sh -c <cmd>` when the `SHELL` env var is unset, which is the default on Windows. Reported in #686.

This PR adds a Windows-specific fallback to `cmd.exe /C <cmd>` when `SHELL` is unset, while preserving the existing `$SHELL -c <cmd>` path whenever `SHELL` is set (so Linux/macOS users, plus Git Bash / WSL / custom-shell users on Windows, are unaffected).

### Why a separate package

`shell.Command(...)` is extracted into a new leaf package `internal/shell` so it can be unit-tested without pulling in the `bubbletea`/`lipgloss` runtime that `internal/tui` depends on. Resolves a testability problem and keeps the resolver as a single, reusable concern.

### Resolution logic

| Condition | Resolved command |
|---|---|
| `SHELL` env set | `$SHELL -c <cmd>` (POSIX convention) |
| Windows + `SHELL` unset | `%COMSPEC% /C <cmd>` (falls back to `cmd.exe` if `COMSPEC` is empty) |
| Otherwise | `sh -c <cmd>` (pre-fix POSIX fallback) |

Non-Windows behavior reduces byte-for-byte to the original logic — strict superset.

### Tests

`internal/shell/shell_test.go` covers four scenarios:

- `TestCommand_HonorsShellEnv` — `SHELL` always wins, regardless of platform.
- `TestCommand_FallbackWithoutShell` — Windows takes the `cmd.exe /C` branch; POSIX takes the `sh -c` branch.
- `TestCommand_HonorsComspecOnWindows` — custom `COMSPEC` paths are respected (Windows-only; skipped elsewhere).
- `TestCommand_PassesCommandStringIntact` — multi-token / quoted command strings are passed through unchanged.

### Empirical verification

Ran on both platforms before opening this PR:

| Platform | `SHELL` | Resolved | Result |
|---|---|---|---|
| Windows 11 | unset (#686 case) | `C:\WINDOWS\system32\cmd.exe /C echo …` | `stdout="…\r\n"`, exit 0 |
| Windows 11 | `/usr/bin/bash` (Git Bash) | `…\Git\usr\bin\bash.exe -c echo …` | `stdout="…\n"`, exit 0 |
| Linux (golang:1.26 container) | unset | `sh -c echo …` | `stdout="…\n"`, exit 0 |

Unit tests: 4/4 PASS on Windows; 3/4 PASS + 1 SKIP (Windows-only test correctly skipped) on Linux.

### Out of scope

- Reporter mentioned wanting a built-in "squash and merge" keybinding — that's a feature request, not a bug fix, and a separate concern from making custom commands work on Windows. Happy to send a follow-up if there's interest, but this PR is focused on closing the crash.
- The `executeCustomCommand` callers (`runCustomPRCommand`, `runCustomBranchCommand`, etc.) are unchanged — only the resolver they share is updated.

Reported by @ES0T.


dlvhdr (approved):
LGTM!

Issue #679, opened 2025-10-24
[BUG] Scrolling the issue list causes display corruption


Nevon:
**Describe the bug**
When scrolling down the list of issues, the display gets all jumbled up. I can reproduce this in multiple terminal emulators, including Ghostty and the built-in mac Terminal app.

Perhaps related to #584, but the reproduction steps are different

**To Reproduce**
1. Have a number of pull requests assigned to you. I had 25 when testing. I'm not sure, but it seems like you need more than just a few. It's likely the same thing would happen with issues, but I can't confirm that for sure.
2. Select the Needs My Review tab and see that it looks as expected (first screenshot)
3. Scroll down the list
4. See the display gets all jumbled up (second screenshot).

**Expected behavior**
I guess it's pretty clear what to expect.

**Screenshots**
I have to redact the content, but hopefully you can see that the top part of the screen gets weirdly duplicated. The same thing happens with the bottom bar as well as the list contents. You also can no longer see the top tabs. The blank areas are from me redacting the contents.

Initial view looks fine:

<img width="2063" height="1324" alt="Image" src="https://github.com/user-attachments/assets/38d89a0f-2966-4b62-866f-628390e547be" />

After scrolling:

<img width="2061" height="1364" alt="Image" src="https://github.com/user-attachments/assets/a669096f-5b1c-47b1-8b0d-3cbd9cf214c0" />

**Desktop (please complete the following information):**
 - OS: Mac
 - Terminal: Ghosttty & Terminal.app
 - Shell: zsh
 - Extension version: v4.18.0 & v4.19.0

Nevon:
Turns out I can reproduce this with my personal Github account as well, where i don't have to redact anything. I cannot reproduce it in the "Needs My Review" tab where I only have 3 PRs assigned, but I can reproduce it in the "My Pull Requests" tab with 6 open PRs.

<img width="2559" height="1364" alt="Image" src="https://github.com/user-attachments/assets/0ec27763-3635-4c64-90b3-9af498281af9" />

reallyely:
Seeing this also. Navigating up and down using `j` `k` causes the the display to stack certain elements like this. 

Using Arch, Alacritty, Bash

Debug.log

```
12:12PM DEBU <cmd/root.go:89> Logging to debug.log
12:12PM DEBU <cmd/root.go:91> Running in repo repo=/home/rowvir/dev/redacted
12:12PM DEBU <tui/ui.go:728> window size changed width=239 height=89
[...]

Issue #584, opened 2025-07-07
[BUG] Display corruption


ShintaroOba:
**Describe the bug**
- Initial (This screen looks fine.)

![Image](https://github.com/user-attachments/assets/dcdb12bd-8ef5-41a3-8b79-3b4e1d42cce2)

- switch the tab (Tabs at the top are not displayed.)

![Image](https://github.com/user-attachments/assets/fccf81fc-7124-47bb-abb3-488981185630)

- When scrolling up and down, the highlighting also seems to behave strangely
![Image](https://github.com/user-attachments/assets/8114ebb0-31f9-462c-915d-0e2ae7708aa3)

**To Reproduce**
1. execute ``gh dash``

**Expected behavior**
To work the same way as demo.

**Desktop (please complete the following information):**
 - OS: Win11

namelessjon:
I also see this on MacOS using iTerm2

winston-riley-zocdoc:
Im seeing this frequently too on MacOS using Terminal and Ghostty

dlvhdr:
I'm not seeing this. Can you provide a repro config and screenshots?

winston-riley-zocdoc:
@dlvhdr this is what Im seeing: https://github.com/charmbracelet/bubbletea/issues/1459#issuecomment-3145357428

its intermittent. but happens frequently. 

Stock config, no modifications to the defaults.


edit: not sure if this is useful but it could be coming from bubble tea: this issue sounds similar: https://github.com/charmbracelet/bubbletea/issues/1459

winston-riley-zocdoc:
@dlvhdr bumping this thread. just making sure you saw the above screen grab.

PowerSchill:
I might be having the same problem but your image has already expired. 

> [@dlvhdr](https://github.com/dlvhdr) this is what Im seeing: https://asciinema.org/a/4zu2M0Aal8KdvGK5QvMPo2LhL
> 
> its intermittent. but happens frequently.
> 
> Stock config, no modifications to the defaults.
> 
> edit: not sure if this is useful but it could be coming from bubble tea: this issue sounds similar: [charmbracelet/bubbletea#1459](https://github.com/charmbracelet/bubbletea/issues/1459)

I might be having the same problem but your image has already expired.

winston-riley-zocdoc:
@PowerSchill here is link of a screen shot: https://github.com/charmbracelet/bubbletea/issues/1459#issuecomment-3145357428

[...]

PR #861, opened 2026-05-04
fix: discard browser launcher stdout/stderr to prevent TUI display corruption (closes #829)

### LINKED ISSUE #829: [BUG] Opening to a browser corrupts display
**Describe the bug**

This is an issue I've had between two different terminal apps. I thought it was the app, so switched to Ghostty (which the docs here use and seem to recommend). But it's still there. Every *single* time I open an issue or PR in the browse (e.g. with the key "o"), it causes the display to corrupt/get jagged. See the screenshot below. This is the same as reported in several other issues e.g. #679 and #584. 

**To Reproduce**

1. Open with `gh dash`
2. In either PR or Issues panel (any words), and "open in GitHub" (key "o").
3. After browser is opened, switch back to terminal app and the display is jagged. Happens every single time.

When I run `gh dash --debug`, I don't see anything in the log. This is the relevant location:

```text
1:46PM INFO <tui/ui.go:645> Task finished id="fetching_prs_3_2026-04-09 13:46:41.403847209 +0200 CEST m=+0.022791348"
1:46PM INFO <tui/ui.go:179> Key pressed key=o
1:46PM INFO <tui/ui.go:85> Starting task id=open_browser_1775735204
1:46PM INFO <tui/ui.go:645> Task finished id=open_browser_1775735204
1:46PM INFO <data/prapi.go:568> Successfully fetched PR url=https://github.com/seedcase-project/check-datapackage/pull/323
1:46PM INFO <tui/ui.go:179> Key pressed key=down
```

I have nothing in my config that modifies the display, its only setting the filters for tabs.

**Expected behavior**

The display should not get corrupted.

**Screenshots**

Image

- OS: Ubuntu 24.04
- Terminal Emulator: Ghostty
- Using tmux? No
- dash version 4.23.2

Booyaka101:
### Summary

All three `browser.New` call sites in gh-dash pass `os.Stdin` as the third argument:

```go
b := browser.New("", os.Stdout, os.Stdin)
//                              ^^^^^^^^^
```

The signature in `cli/go-gh/v2/pkg/browser` is `New(launcher string, stdout, stderr io.Writer) *Browser`. The third argument should be a stderr writer. It compiles because `*os.File` implements `io.Writer`, but the launcher subprocess (`xdg-open`, `gnome-open`, `start`) inherits the TTY, and any stderr it emits — GTK / GVFS warnings, GNOME deprecation messages — gets routed through that fd onto the same TTY the TUI is rendering to. That's the display corruption reported in #829, and matches the symptoms in the related #584 and #679 threads.

PR #569 added a manual `Ctrl+L` redraw keybinding as a mitigation, but the leak source was never addressed.

### Fix

Pass `io.Discard` for both writers — for a TUI we never want launcher noise reaching the terminal. Three call sites:

```diff
-		b := browser.New("", os.Stdout, os.Stdin)
+		// Discard the launcher's stdout/stderr so any noise (e.g. GTK / GVFS
+		// warnings from xdg-open / gnome-open) does not leak into the TUI's
+		// terminal and corrupt the display. See #829, #584, #679.
+		b := browser.New("", io.Discard, io.Discard)
```

- [`internal/tui/tasks.go:28`](internal/tui/tasks.go#L28) — main "open in browser" task
- [`internal/tui/components/notificationssection/commands.go:255`](internal/tui/components/notificationssection/commands.go#L255) — notification-row open
- [`internal/tui/ui.go:816`](internal/tui/ui.go#L816) — donate button

`cli/cli` itself uses the analogous pattern: [`browser.New("", io.Out, io.ErrOut)`](https://github.com/cli/cli/blob/trunk/pkg/cmd/factory/default.go) backed by an iostreams wrapper. Discarding here is more aggressive than reproducing iostreams — for a full-screen TUI the launcher's output is never wanted.

### Build verification

| Platform | Result |
|---|---|
| Windows + Go 1.26.2 | `go build ./...` → exit 0 |
| Linux + Go 1.26 (Docker `golang:1.26`) | `go build ./...` → PASS |

I haven't run the `internal/tui` test suite because it hangs on package init (charm/bubbletea TTY setup) on this Windows environment — same issue I hit on #686. The change is purely an import + 3-line callsite swap; build green on both platforms covers it.

### Out of scope

- The `Ctrl+L` redraw keybinding from #569 stays useful for any other source of corruption. This PR removes the most common one (browser launcher noise) but doesn't replace the redraw escape hatch.
- I haven't added a test exercising the actual launcher subprocess — verifying the writers are discarded would require either intercepting the `*Browser` constructor or running an end-to-end TUI test, both of which are heavier than the fix justifies.

Reported by @lwjohnst86, @ShintaroOba, @404Simon.

### DIFF
--- a/internal/tui/components/notificationssection/commands.go
+++ b/internal/tui/components/notificationssection/commands.go
@@ -3,6 +3,7 @@ package notificationssection
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strings"
@@ -252,7 +253,10 @@ func (m *Model) openInBrowser() tea.Cmd {
 			}
 		},
 		func() tea.Msg {
-			b := browser.New("", os.Stdout, os.Stdin)
+			// Discard the launcher's stdout/stderr so any noise (e.g. GTK /
+			// GVFS warnings from xdg-open / gnome-open) does not leak into
+			// the TUI's terminal and corrupt the display. See #829, #584, #679.
+			b := browser.New("", io.Discard, io.Discard)
 			err := b.Browse(notificationUrl)
 			if err != nil {
 				return constants.ErrMsg{Err: err}
--- a/internal/tui/tasks.go
+++ b/internal/tui/tasks.go
@@ -3,7 +3,7 @@ package tui
 import (
 	"errors"
 	"fmt"
-	"os"
+	"io"
 	"reflect"
 	"time"
 
@@ -25,7 +25,10 @@ func (m *Model) openBrowser() tea.Cmd {
 	}
 	startCmd := m.ctx.StartTask(task)
 	openCmd := func() tea.Msg {
-		b := browser.New("", os.Stdout, os.Stdin)
+		// Discard the launcher's stdout/stderr so any noise (e.g. GTK / GVFS
+		// warnings from xdg-open / gnome-open) does not leak into the TUI's
+		// terminal and corrupt the display. See #829, #584, #679.
+		b := browser.New("", io.Discard, io.Discard)
 		currRow := m.getCurrRowData()
 		if currRow == nil || reflect.ValueOf(currRow).IsNil() {
 			return constants.TaskFinishedMsg{
--- a/internal/tui/ui.go
+++ b/internal/tui/ui.go
@@ -2,6 +2,7 @@ package tui
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"reflect"
 	"runtime/debug"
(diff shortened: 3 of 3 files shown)


dlvhdr:
@dlvlvlvlvl

Booyaka101:
That looks like it may have been a typo or cut-off — `@dlvlvlvlvl` doesn't resolve to a user. Was that meant to be `@` someone in particular, or any other feedback I should address on this one?

PR #868, opened 2026-05-09
Default confirmation prompts to No for destructive actions

### LINKED ISSUE #596: Default option for close/reopen PR/issue should be "no" (safer)
It feels unsafe to have the default be "Y", especially in situations where <kbd>x</kbd>/<kbd>X</kbd> is pressed accidentally (e.g., when the user was expecting to be operating on a different window, see issue #595).

The default option for closing/reopening should be "N". This also makes gh-dash more compliant to popular good practices like human error prevention, conservative/defensive design, and even the Principle of Least Astonishment (POLA).

None

**Additional context**
None

kimjune01:
Fixes #596.

Close/reopen/delete prompts default to Yes — pressing Enter without
typing triggers the action. This is unsafe for destructive operations.

Flip all confirmation prompts from `(Y/n)` to `(y/N)` and remove the
`input == ""` accept condition so empty input no longer confirms.
Also fixes an inconsistency in notification view where the prompt
showed `(y/N)` but Enter still confirmed.

### DIFF
--- a/internal/tui/components/issuessection/issuessection.go
+++ b/internal/tui/components/issuessection/issuessection.go
@@ -87,7 +87,7 @@ func (m *Model) Update(msg tea.Msg) (section.Section, tea.Cmd) {
 			case "enter":
 				input := m.PromptConfirmationBox.Value()
 				action := m.GetPromptConfirmationAction()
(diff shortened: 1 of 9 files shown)


dlvhdr (approved):
LGTM

PR #873, opened 2026-05-13
Add notification cursor regression coverage

### LINKED ISSUE #858: For notifications, after marking the last item as done keep the cursor on the new last item
For notifications, we can mark them as done with `D`.
When marked as done, they are removed from the view.
In this occasion, the cursor (current row) stays in place, meaning the notification right below will become selected.
This behavior is fine, but the edge case is when we are on the last item.
In this case, we will have no row / item selected no more which means we cannot mark multiple items in succession as done from the bottom up.
This I consider to be a common usecase since you often want to process notifications starting with the oldest ones.

In the case the last item is marked as done, the new last item should be selected afterwards.

NA

**Additional context**
NA

Kyle-Structure:
## Summary
- add a regression test for removing the selected last notification
- assert the notifications section keeps the cursor on the new last item after the row is removed

Fixes #858

## Verification
- go test ./internal/tui/components/notificationssection
- git diff --check
- git diff --cached --check


dlvhdr (approved):
LGTM

PR #891, opened 2026-05-19
feat(prview): PR review-comment threading, diff hunks, and web-parity activity (closes #852)

### LINKED ISSUE #852: PR activity view flattens review threads and drops the code context

ysrfz:
# Summary

- [x] Closes issue #852
- [x] I have read the [CONTIBUTING.md](../CONTRIBUTING.md) and [AI_POLICY.md](../AI_POLICY.md) guides

**Status: WIP — please comment if you don't like the current direction while I polish the edges.**

Reworks the PR activity panel to look more like GitHub web's Conversation tab. A few things going on here:

- **Thread grouping** — review comments now sit under the thread they belong to instead of flat-by-timestamp. Replies get a compact `↳ {author}` prefix.
- **Diff hunks** — fetch `diffHunk` on review comments and render the last 4 content lines above each thread (mirroring the collapsed preview on web). `@@` headers get stripped since they're just positional metadata.
- **Thread badges** — `[Outdated]` and `[Resolved]` render inline next to the anchor, and wrap to the next line when the anchor is too wide rather than getting clipped.
- **DISMISSED glyph** — was rendering as blank space before. Now uses a faint `↺` so dismissed approvals don't look like rendering bugs.
- **Implicit-Review unwrap** — GitHub auto-generates a `state=COMMENTED, body=""` wrapper whenever a thread starts outside a formal review. The TUI was rendering these as empty `xyz reviewed` blocks. Now: 0 threads → hide; 1 → promote the thread up a level; 2+ → promote them all as siblings.
- **Conversation count** — the `N comments` figure in the title now matches the web `(N)` badge, including the slightly weird case where reviews with `state=COMMENTED` + body + zero inline comments get excluded (web treats them as redundant with plain IssueComments, even though it still renders the card — I went back and forth on this one, ended up matching web for consistency).

Keybindings and overall panel layout unchanged.

- `go test ./internal/tui/components/prview ./internal/data` — added 14 new tests in `activity_test.go` covering each behaviour above (thread grouping, diff-hunk truncation, badges, DISMISSED glyph, implicit-review unwrap cases, count-badge web-parity)
- `go build && go vet ./...` clean
- Manual side-by-side against `gh-dash` v4.24.1 on a 20+-review PR (mixed thread/IssueComment/DISMISSED activity); will attach screenshots once the UI is polished

## Images/Videos

(will attach side-by-side screenshots once I'm happy with the UI)

## AI Usage Disclosure

Per [AI_POLICY.md](../AI_POLICY.md): wrote the code with **Claude Code** as a pair-programming assistant, with substantial AI involvement on the implementation drafts and tests. Each design decision (web-parity counting, implicit-review unwrap cases, diff-hunk truncation length, badge wrap behaviour) was driven and reviewed by me, with multiple rounds of pushback on the approach before landing. I understand the changes and can explain how the activity-panel pipeline interacts with the GraphQL fragment and the existing PR view — happy to walk through any part on request.

Closes #852

### DIFF
--- a/internal/data/prapi.go
+++ b/internal/data/prapi.go
@@ -241,6 +241,7 @@ type Comment struct {
 		Login string
 	}
 	Body      string
+	CreatedAt time.Time
 	UpdatedAt time.Time
 }
 
@@ -248,10 +249,18 @@ type ReviewComment struct {
 	Author struct {
 		Login string
 	}
-	Body      string
-	UpdatedAt time.Time
-	StartLine int
-	Line      int
+	Body              string
+	CreatedAt         time.Time
+	UpdatedAt         time.Time
+	StartLine         int
+	Line              int
+	DiffHunk          string
+	PullRequestReview struct {
+		Id string
+	}
+	ReplyTo struct {
+		Id string
+	}
 }
 
 type ReviewComments struct {
@@ -268,12 +277,14 @@ type ReviewThreads struct {
 }
 
 type Review struct {
+	Id     string
 	Author struct {
 		Login string
 	}
-	Body      string
-	State     string
-	UpdatedAt time.Time
+	Body        string
+	State       string
+	SubmittedAt time.Time
+	UpdatedAt   time.Time
 }
 
 type Reviews struct {
@@ -285,6 +296,7 @@ type ReviewThreadsWithComments struct {
 	Nodes []struct {
 		Id           string
 		IsOutdated   bool
+		IsResolved   bool
 		OriginalLine int
 		StartLine    int
 		Line         int
--- a/internal/tui/components/prview/activity.go
+++ b/internal/tui/components/prview/activity.go
@@ -3,6 +3,7 @@ package prview
 import (
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"charm.land/glamour/v2"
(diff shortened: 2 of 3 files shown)


dlvhdr:
Hey @ysrfz :)
I've updated my AI policy to disallow any AI usage.
I am willing to explore your idea though and I think a first good step would be to open an issue to discuss it.
We can then make sure to take performance into account as well as build a figma mock we can agree on.
wdyt?

Nadzkielegit (approved):

PR #894, opened 2026-05-21
fix: use shared-state background color to avoid nil pointer (#876)

### LINKED ISSUE #876: [BUG] TUI crashes with nil pointer dereference when BackgroundColorMsg not reached
**Describe the bug**
Program panics when attempting to render markdown content (right side pane) due to nil pointer dereference.
Fatal error happens in `gh-dash/internal/tui/markdown/markdownRenderer.go` (`GetMarkdownRenderer` attempts to dereference `markdownStyle` without prior check).
This condition occurs when the terminal does not respond to the `tea.RequestBackgroundColor` command. When that happens, `case tea.BackgroundColorMsg` is never reached, leaving `markdownStyle` uninitialized.

I'd be happy to contribute a PR. I've got a working concept [here](https://github.com/fsorodrigues/gh-dash/tree/fix/md-style-nil-pointer-dereference), but also cognizant the maintainers may have other ideas re supporting unconventional setups or how to solve this issue.

**To Reproduce**

1. Start `gh dash` from a repo that contains PRs or issue
2. Trigger search so that the program returns a list of PRs or issues. This can sometimes be the initial search state. I've done so by altering the filters in `config.yaml`
3. The program panics and exits

**Expected behavior**
The program should render the contents of PR/issue in markdown on the right side pane or, at least, recover to a graceful state or graceful exit.

**Screenshots**
[panic output](https://github.com/user-attachments/files/27788134/panic.txt)
[debug logs](https://github.com/user-attachments/files/27788401/debug.log)
https://github.com/user-attachments/assets/ff063929-af60-484a-84ab-cd995e29c2e6

- OS: WSL (Ubuntu 24.04 on Windows 11)
- Terminal Emulator: Alacritty
- Using tmux? Yep 🙈 , but issue also occurs when tmux detached
- Vesion: 4.24.1

fsorodrigues:
## Summary
Fixes #876 — TUI crashes with nil pointer dereference of `markdownStyle` when
`BackgroundColorMsg` not reached

This PR introduces background color shared state, default and fallback behaviors, and changes to the `InitializeMarkdownStyle` and `GetMarkdownRender` functions signatures to ensure that `*markdownStyle` is never nil when dereferenced.

- [x] Closes issue #876
- [x] I have read the [CONTIBUTING.md](https://github.com/dlvhdr/gh-dash/CONTRIBUTING.md) and [AI_POLICY.md](https://github.com/dlvhdr/gh-dash/AI_POLICY.md) guides

## AI disclosure (per AI_POLICY.md)
Diagnosis of nil pointer dereference location was done with assistance of GPT-5.4 (via OpenCode). Same AI was also used to identify `GetMarkdownRender` usage extense given the bugfix plan I'm proposing involved altering the function's signature.

## Solution
Here are the changes introduced:

`internal/tui/context/context.go`
- introduce 2 new fields to the `ProgramContext` struct: `HasDarkBackground` bool, `BackgroundSource` string

`interna/tui/markdown/markdownRenderer.go`
- alter `InitializeMarkdownStyle` and `GetMarkdownRenderer` signatures to include `*context.ProgramContext`. These ensure the shared state values `ctx.HasBackground` and `ctx.BackgroundSource` are available for the create of the markdown renderer.

`internal/tui/ui.go`
- initialize program context with defaults for `ctx.HasBackground` and `ctx.BackgroundSource`
- create fallback behavior that uses lipgloss's `compat.HasDarkBackground` as boolean for `ctx.HasDarkBackground`
- set background color fields in context from `tea.BackgroundColorMsg` when reached
- debug log statements with background color values

`internal/tui/components/issueview/activity.go`
`internal/tui/components/issueview/issueview.go`
`internal/tui/components/prview/activity.go`
`internal/tui/components/prview/prview.go`
`internal/tui/modelUtils.go`
- call `GetMarkdownRender` with context as argument

## Testing
Verified the nil pointer derefence panic present on tag v4.24.1 on setup described in #876.

Rebuilt binary with patch (commit 46b5843787fc), removed and reinstalled dash extension from local repo

Started gh dash with debug flag. Verified nil pointer derefence panic is gone and that debug logs describe shared state as expected. 

Verified gh dash works normally and renders markdown contents of PRs and issues with expected background

### DIFF
--- a/internal/tui/markdown/markdownRenderer.go
+++ b/internal/tui/markdown/markdownRenderer.go
@@ -4,14 +4,19 @@ import (
 	"charm.land/glamour/v2"
 	"charm.land/glamour/v2/ansi"
 	"charm.land/glamour/v2/styles"
+	log "charm.land/log/v2"
 
 	"github.com/dlvhdr/gh-dash/v4/internal/tui/context"
 )
 
-var markdownStyle *ansi.StyleConfig
+var (
+	markdownStyle       *ansi.StyleConfig
+	markdownStyleSource string
+)
 
 func InitializeMarkdownStyle(ctx *context.ProgramContext) {
-	if markdownStyle != nil && ctx.BackgroundSource == "bubbletea" {
+	if markdownStyle != nil && markdownStyleSource == "bubbletea" {
+		log.Debugf("InitializeMarkdownStyle: keeping existing bubbletea style")
 		return
 	}
 
@@ -20,6 +25,13 @@ func InitializeMarkdownStyle(ctx *context.ProgramContext) {
 	} else {
 		markdownStyle = &styles.LightStyleConfig
 	}
+	markdownStyleSource = ctx.BackgroundSource
+
+	log.Debugf(
+		"InitializeMarkdownStyle: assigned ctx.hasDarkBackground=%t, markdownStyleSource=%q",
+		ctx.HasDarkBackground,
+		markdownStyleSource,
+	)
 }
 
 func GetMarkdownRenderer(width int, ctx *context.ProgramContext) glamour.TermRenderer {
--- a/internal/tui/markdown/markdownRenderer.go
+++ b/internal/tui/markdown/markdownRenderer.go
@@ -4,22 +4,29 @@ import (
 	"charm.land/glamour/v2"
 	"charm.land/glamour/v2/ansi"
 	"charm.land/glamour/v2/styles"
+
+	"github.com/dlvhdr/gh-dash/v4/internal/tui/context"
 )
 
 var markdownStyle *ansi.StyleConfig
 
-func InitializeMarkdownStyle(hasDarkBackground bool) {
-	if markdownStyle != nil {
+func InitializeMarkdownStyle(ctx *context.ProgramContext) {
+	if markdownStyle != nil && ctx.BackgroundSource == "bubbletea" {
 		return
 	}
-	if hasDarkBackground {
+
+	if ctx.HasDarkBackground {
 		markdownStyle = &CustomDarkStyleConfig
 	} else {
 		markdownStyle = &styles.LightStyleConfig
 	}
 }
 
-func GetMarkdownRenderer(width int) glamour.TermRenderer {
+func GetMarkdownRenderer(width int, ctx *context.ProgramContext) glamour.TermRenderer {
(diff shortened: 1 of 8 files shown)


internal/tui/markdown/markdownRenderer.go
```diff
@@ -4,22 +4,29 @@ import (
...
+
+	"github.com/dlvhdr/gh-dash/v4/internal/tui/context"
 )
 
 var markdownStyle *ansi.StyleConfig
 
-func InitializeMarkdownStyle(hasDarkBackground bool) {
-	if markdownStyle != nil {
+func InitializeMarkdownStyle(ctx *context.ProgramContext) {
+	if markdownStyle != nil && ctx.BackgroundSource == "bubbletea" {
```
dlvhdr: I am not sure how this is supposed to work, but wouldn't we want to use the bubbletea background source when the `BackgroundColorMsg` arrives?
So I'd assume that we'd want to save the background source in a global var next to `markdownStyle` to know which source initialized it and if it's not nil but the source is `compat` - initialize it with the new `bubbletea` source.
Correct me if I'm wrong..

fsorodrigues: so... what I had in mind is a progression of background color "certainty"
- in tui/ui.go we set a default state for dark bg /  dark bg source when the Model and Context are created. I when with a hardcoded true, which I'm happy for y'all to do whatever initial state you think is right.
- then in InitMsg we can "upgrade" the background color data in the Context using the value from lipgloss compat, which won't be a hardcoded value.
- finally, if/when the tea.BackgroundColorMsg arrives, we have the "most certain" background color to set the style with "finality". The finality being the reason for that addition to the if condition. However, I do see, now, that I made the bubbletea route effectively dead on arrival 🫤 

I'll give it another crack in the morning with your suggestion.

fsorodrigues: @dlvhdr I think you can have another look.
On the setup that was crashing (wsl + alacritty + tmux), I now see the style initialized with the compat source.
On a setup that never saw the crash (macOs + ghostty + tmux), it follows the expected progression and reaches the bubbletea source.

dlvhdr:
this looks great @fsorodrigues, thank you!

dlvhdr (approved):

PR #895, opened 2026-05-21
fix: accept pgdn keybinding alias

### LINKED ISSUE #465: [BUG] Can't map preview page up/down to pgup/pgdn keys
**Describe the bug**
Mapping `pgup` to `builtin: pageUp` and `pgdn` to `builtin: pageDown` has no effect, the keys do nothing. Mapping them to something else works, so the problem is not in receiving the key events.

**To Reproduce**
1. Add this to your config:
```yaml
keybindings:
  universal:
    - key: pgup
      builtin: pageUp
    - key: pgdn
      builtin: pageDown
```
2. Open `gh dash` and try pressing pgup/pgdown

**Expected behavior**
The preview pane scrolls up and down

 - OS: Manjaro Linux

**Side note**
It was difficult to find the list of supported keys and their string representations. In the end I tracked it down [here](https://github.com/charmbracelet/bubbletea/blob/6e71f52a8add0fdeba202d4e1bdd289182b156ac/key.go#L260-L350) by looking through the code.

puneetdixit200:
## Summary
- [x] Closes issue #465
- [x] I have read the [CONTIBUTING.md](../CONTRIBUTING.md) and [AI_POLICY.md](../AI_POLICY.md) guides

- normalize configured `pgdn` keybindings to Bubble Tea's `pgdown` key event name
- preserve the configured help label so help still shows `pgdn`
- add regression coverage for rebinding `pageDown` to `pgdn`

Fixes #465

## AI assistance
OpenAI GPT-5 assisted with drafting the patch and verification steps. I reviewed the change, checked it against the keybinding behavior, and kept responsibility for the final submission.

Red test before the fix:
- `go test ./internal/tui/keys -run TestRebindUniversalPageDownAcceptsPgdnAlias -count=1 -v` failed because the PageDown event string was `pgdown` and did not match configured `pgdn`

After the fix:
- `go test ./internal/tui/keys -run TestRebindUniversalPageDownAcceptsPgdnAlias -count=1 -v`
- `go test ./internal/tui/keys -count=1`
- `go test ./internal/tui -run 'Test.*Keybinding|Test.*Page|Test.*Preview' -count=1`
- `go test ./... -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --path-mode=abs --config=.golangci.yml --timeout=5m ./internal/tui/keys`
- `git diff --check`

## Images/Videos
Not applicable.

### DIFF
--- a/internal/tui/keys/keys.go
+++ b/internal/tui/keys/keys.go
@@ -245,6 +245,15 @@ var (
 	CustomNotificationBindings []key.Binding
 )
 
+func normalizeConfigKey(key string) string {
+	switch key {
+	case "pgdn":
+		return "pgdown"
+	default:
+		return key
+	}
+}
+
 func rebindUniversal(universal []config.Keybinding) error {
 	log.Debug("Rebinding universal keys", "keys", universal)
 
@@ -260,7 +269,7 @@ func rebindUniversal(universal []config.Keybinding) error {
 				}
 
 				customBinding := key.NewBinding(
-					key.WithKeys(kb.Key),
+					key.WithKeys(normalizeConfigKey(kb.Key)),
 					key.WithHelp(kb.Key, name),
 				)
 
@@ -316,7 +325,7 @@ func rebindUniversal(universal []config.Keybinding) error {
 			return fmt.Errorf("unknown built-in universal key: '%s'", kb.Builtin)
 		}
 
-		key.SetKeys(kb.Key)
+		key.SetKeys(normalizeConfigKey(kb.Key))
 
 		helpDesc := key.Help().Desc
 		if kb.Name != "" {
--- a/internal/tui/keys/keys_test.go
+++ b/internal/tui/keys/keys_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"charm.land/bubbles/v2/key"
+	tea "charm.land/bubbletea/v2"
 
 	"github.com/dlvhdr/gh-dash/v4/internal/config"
 )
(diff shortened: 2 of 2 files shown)


dlvhdr:
Not sure I understand @puneetdixit200 - did `pgdn` used to work but it changed to `pgdown` in bubbletea v2?

puneetdixit200:
I don't know whether `pgdn` worked before the Bubble Tea v2 update. The issue here is that the current key event string for PageDown is `pgdown`, while the reported config uses the common shorter alias `pgdn`, so the configured binding is registered under a string that does not match the emitted key event.

This patch keeps the user's configured/help text as `pgdn`, but registers `pgdown` internally for matching. `pgup` already matches the current Bubble Tea spelling, so it does not need the same alias mapping.

dlvhdr:
not sure I understand. Why wouldn't the user just pass the correct key then? he can pass `pgdown` directly so why do we require him to pass `pgdn` and then translate that?

puneetdixit200:
`pgdown` already works and this PR does not require users to use `pgdn`. The reason I added the alias is compatibility with the issue report and with the common PgDn spelling users tend to put in config files.

So the behavior after this patch is:
- `pgdown` remains the canonical Bubble Tea event name and continues to work
- `pgdn` is accepted as an alias at config-load time
- help text still shows the key the user configured

If you prefer to keep config values limited to the exact Bubble Tea names, I am fine closing this and leaving the issue as a documentation/config correction instead.

dlvhdr:
I prefer to limit the options to bubbletea keys. I would like to document that however, in the docs site. A contribution around that would be awesome

Issue #922, opened 2026-06-25
[BUG] markdown.GetMarkdownRenderer(_) panic


travelliu:
runtime error: invalid memory address or nil pointer dereference

Restoring terminal...

goroutine 1 [running]:
runtime/debug.Stack()
        /opt/hostedtoolcache/go/1.25.9/x64/src/runtime/debug/stack.go:26 +0x5e
charm.land/bubbletea/v2.(*Program).recoverFromPanic(0xc000446000, {0xd1e4a0, 0x1930030})
        /home/runner/go/pkg/mod/charm.land/bubbletea/v2@v2.0.2/tea.go:1260 +0x152
charm.land/bubbletea/v2.(*Program).Run.func2()
        /home/runner/go/pkg/mod/charm.land/bubbletea/v2@v2.0.2/tea.go:1011 +0xe6
panic({0xd1e4a0?, 0x1930030?})
        /opt/hostedtoolcache/go/1.25.9/x64/src/runtime/panic.go:783 +0x132
github.com/dlvhdr/gh-dash/v4/internal/tui/markdown.GetMarkdownRenderer(_)
        /home/runner/go/pkg/mod/github.com/dlvhdr/gh-dash/v4@v4.24.1/internal/tui/markdown/markdownRenderer.go:24 +0x41
github.com/dlvhdr/gh-dash/v4/internal/tui/components/issueview.(*Model).renderBody(0xc001bcee40)
        /home/runner/go/pkg/mod/github.com/dlvhdr/gh-dash/v4@v4.24.1/internal/tui/components/issueview/issueview.go:228 +0xb2
github.com/dlvhdr/gh-dash/v4/internal/tui/components/issueview.Model.View({0xc00055c000, 0xc00102c000, 0x1, 0x53, {0xc00055c000, {0xc00055c000, 0xc0004fca08, 0x0, {0x0, {...}, ...}, ...}, ...}})
        /home/runner/go/pkg/mod/github.com/dlvhdr/gh-dash/v4@v4.24.1/internal/tui/components/issueview/issueview.go:144 +0x970
github.com/dlvhdr/gh-dash/v4/internal/tui.(*Model).syncSidebar(0xc0005f2008)
        /home/runner/go/pkg/mod/github.com/dlvhdr/gh-dash/v4@v4.24.1/internal/tui/ui.go:1239 +0x9c5
[...]

PR #924, opened 2026-06-30
fix(markdown): handle nil ProgramContext in GetMarkdownRenderer

AruneshDwivedi:
- [x] Closes issue #922

GetMarkdownRenderer dereferenced ctx without a nil check, causing a panic when the issue view is rendered before the program context is initialized. Falls back to a default renderer (no custom styling) when ctx is nil.


AruneshDwivedi:
Closing — this project has a strict No-AI Policy and I respect that boundary.

PR #930, opened 2026-07-08
fix: remove visible logs when not in debug mode

dlvhdr:
# Summary

- [] Closes issue #...
- [] I have read the [Contributing](../CONTRIBUTING.md) and the strict [No-AI Policy](../AI_POLICY.md) guides.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -186,7 +186,7 @@ func init() {
 	)
 
 	rootCmd.Run = func(_ *cobra.Command, args []string) {
-		debug, err := rootCmd.Flags().GetBool("debug")
+		debug, _ := rootCmd.Flags().GetBool("debug")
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -245,21 +237,16 @@ func init() {
 			log.Warn("did not find github repo at current path")
 		}
 
-		debug, err := rootCmd.Flags().GetBool("debug")
 		if err != nil {
 			log.Fatal("Cannot parse debug flag", err)
 		}
```
PatrickTulskie, line 242: Looks like these need to get deleted too. This check is now incorrectly fatally crashing if we're not in a git repo since the err they originally were supposed to crash on got moved and discarded to line 189. That should just fix https://github.com/dlvhdr/gh-dash/issues/932

PR #941, opened 2026-07-23
fix: correct dateFormat schema type

Nick2bad4u:
## Summary

- correct `defaults.dateFormat` from an integer to a string in the published configuration schema
- align the schema default with the documented and runtime default, `relative`

## Evidence

`internal/config/parser.go` defines `DateFormat string`, and the configuration documentation describes either `relative` or a Go time format string. The current schema mistakenly copied the integer shape/default from `refetchIntervalMinutes`, causing valid configurations to fail validation.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm build`

The docs build succeeds. It reports the same three existing broken internal links unrelated to this change.

### DIFF
--- a/docs/src/pages/schema/defaults.json.ts
+++ b/docs/src/pages/schema/defaults.json.ts
@@ -81,9 +81,8 @@ export function GET() {
         dateFormat: {
           title: "Date format",
           description: "Specifies how dates are formatted.",
-          type: "integer",
-          minimum: 1,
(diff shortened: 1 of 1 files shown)


dlvhdr:
Hey @Nick2bad4u, the other option is to provide a go formatted date. So it is a string but I think it's important to specify its format. Can you add a link to the Go date formats?

PR #957, opened 2026-08-21
fix(search): use advanced search for OR/parentheses support

### LINKED ISSUE #955: Advanced search OR/parentheses in section filters returns no results (works fine via `gh pr list --search`)
### Describe the bug
Using GitHub's advanced search boolean syntax (AND/OR with parentheses) in a section's `filters` config returns no/incorrect results in gh-dash, even though the identical query string works correctly when run directly via `gh pr list --search` or `gh search prs`.

### Steps to reproduce
1. Confirm the query works via plain gh CLI:
   `gh pr list --search 'org:myorg (assignee:@me OR author:@me)'` → returns expected PRs

2. Add the same string as a gh-dash section filter:
```yaml
   prSections:
     - title: "Mine"
       filters: `org:myorg (assignee:@me OR author:@me)`
```
3. Launch gh-dash → section returns no results / incorrect results.

> Or using `/` search and pasting -> same empty result

### Expected behavior
Since gh CLI (v2.79.0+) supports advanced issue/PR search (AND/OR, nested parentheses) via the advanced_search GraphQL/REST parameter, gh-dash's section filters should support the same syntax, since it's documented as passing arbitrary GitHub search qualifiers.

### Suspected cause
gh-dash appears to build its own GraphQL search queries via go-gh rather than shelling out to `gh <cmd> --search`, and may not be requesting the ISSUE_ADVANCED search type/advanced_search=true parameter that GitHub's API needs to parse boolean logic. Without it, the query silently falls back to the classic search engine where OR is treated as a literal token.

### Environment
- gh-dash version:
```
gh-dash version dev
goos: darwin
goarch: arm64
module version: v4.25.2, checksum: h1:i24BUObEJkr0/i1Dt0wEWOBecKMYFqC6J0uQhXTt8NI=
```
- gh version:
```
gh version 2.97.0 (2026-07-31)
https://github.com/cli/cli/releases/tag/v2.97.0
```

erzhtor:
# Summary

Closes #955

GitHub's classic search backend (`type: ISSUE`) treats `OR` and parentheses as literal text, so advanced boolean filters silently return 0/wrong results. GitHub added an advanced search backend [type: ISSUE_ADVANCED](https://github.blog/changelog/2025-03-06-github-issues-projects-api-support-for-issues-advanced-search-and-more/#mag-api-support-for-issues-advanced-search) in March 2025 that parses `AND`/`OR` and nested parentheses. 

The PR switches both the PR and issue search queries to use it.

- [x] I have read the [Contributing](../CONTRIBUTING.md) and the strict [No-AI Policy](../AI_POLICY.md) guides.

Swapped my local `gh dash` extension for this branch's build to confirm the fix end-to-end:
1. `git clone -b fix/advanced-search --single-branch git@github.com:erzhtor/gh-dash.git`
1. `cd gh-dash`
1. `gh extension remove dash`
1. `go build -o gh-dash .`
1. `gh extension install .`
1. Open `gh dash` and try out search/filters using 'OR/AND', e.g. `is:pr (assignee:@me OR author:@me)`


internal/tui/common/labels.go
```diff
@@ -136,7 +136,7 @@ func RenderLabels(labels []data.Label, opts LabelOpts) string {
 		overflowPillStyle := opts.PillStyle.
 			BorderForeground(lipgloss.ANSIColor(244)).
 			Background(lipgloss.ANSIColor(244))
-		if rowBackground := opts.RowStyle.GetBackground(); rowBackground != nil {
+		if rowBackground := opts.RowStyle.GetBackground(); rowBackground != (lipgloss.NoColor{}) {
```
erzhtor, line 139: This seems to an existing issue, golangci-lint was complaining. Updated to get CI green.

erzhtor:
Hey @dlvhdr! Just following up on this, do you think we can get this merged or is there anything you’d like adjusted first?

dlvhdr:
I just haven’t had the time to test it. I did read up on GitHub’s docs site but couldn’t understand why this isn’t the new default..? Isn’t this a superset of the old behavior?

erzhtor:
> I just haven’t had the time to test it. I did read up on GitHub’s docs site but couldn’t understand why this isn’t the new default..? Isn’t this a superset of the old behavior?

Good question, my bad not clarifying. It seems, the main diff is that advanced treats multiple `repo:`, `org:`, or `user:` filters as `OR` unless explicitly used `AND`, whereas old search type treats them as `AND` by default, which is why GH kept advanced search opt-in I guess: https://github.blog/changelog/2025-07-17-duplicate-issues-create-from-anywhere-and-more/

image

Now, I haven’t used multiple repo/user/org filters myself, didn't had a use-case, so I missed it. If this breaking change doesn’t work, let me know I can close the PR and we can keep the current behavior.

dlvhdr:
Oh I understand why then. I think in this case it's quite a big breaking change. I've used queries with multiple `user` filters and that would break it. I wonder if we can add a preference to opt into advanced search on a per section/global level.

